PR #3, opened 2019-05-28
vm-device design codes

liujing2:
This PR brings the basic vm-device design with

- The high level `Device trait` for abstraction and interface.
- A `DeviceManager` keeps track of all registered devices.
- README and test to see details.

Also, it includes vm-allocator crate codes, which provides system resources allocation APIs for other crates.


sameo:
@liujinCould you please add a `.buildkite/pipeline.yml` file to trigger the CI? The buildkite agent stuff has been added so it should work (famous last words...).

liujing2:
@sameo  Sure.

src/device_manager.rs
```diff
@@ -0,0 +1,400 @@
...
+
+/// Simplify the `Result` type.
+pub type Result<T> = result::Result<T, Error>;
+
+/// System device manager serving for all devices management and VM exit handling.
+pub struct DeviceManager<'a> {
+    /// System allocator reference.
+    resource: &'a mut SystemAllocator,
+    /// Devices information mapped by name.
+    devices: HashMap<String, DeviceDescriptor>,
```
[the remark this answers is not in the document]
liujing2: @serban300 For the vmm usage scenario, I think it is reasonable to assume different names for different devices to record the basic device information, which might be used to dump device tree or something like this.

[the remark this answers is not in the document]
chao-p: That's not the case, one example is one device type can have multiple instances, e.g. if you don't include the instance_id as part of name, then those instances will have the same name. With the current Device trait, the device itself may even don't know the instance_id so can't include that when the name() gets called.

I prefer to define a lighter, unique device ID for each device that we can allocate the ID in a center place to guarantee the ID is unique across a VM instance and the program refers the device using this unique ID .

src/device_manager.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    fn remove(&mut self, name: String) -> Option<DeviceDescriptor> {
+        self.devices.remove(&name)
+    }
+
+    fn device_descriptor(
+        &self,
+        dev: Arc<Mutex<dyn Device>>,
+        parent_bus: Option<Arc<Mutex<dyn Device>>>,
+        resource: Vec<IoResource>,
```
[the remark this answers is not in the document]
liujing2: Yep! Will change that.

src/device_manager.rs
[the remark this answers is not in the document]
liujing2: @serban300 Right, previously I though passing a name to e.g. `VirtioBlk::new("virtio-blk-1") ` for each device type to make different instances different names. According to your and @chao-p 's suggestion, I agree that we define a light struct might be like this,
```
struct device_info {
    device_type: String, // Actually a name of type e.g. virtio-blk
    instance_id: u32,
}
```
We allocate a global instance_id to every device instance from number 1, and put this struct into `DeviceDescriptor`, replacing the first member currently. What do you think? :)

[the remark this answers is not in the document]
serban300: Using a structure would be perfect for us. It would be consistent with what we already have in Firecracker.

I think the device_type can be something lighter than a string. For example in Firecracker we use an enum:

```
/// Types of devices that can get attached to this platform.
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub enum DeviceType {
    Virtio(u32),
    #[cfg(target_arch = "aarch64")]
    Serial,
    #[cfg(target_arch = "aarch64")]
    RTC,
}
```
Where the u32 associates to the Virtio can be:
```
/// Types taken from linux/virtio_ids.h.
pub const TYPE_NET: u32 = 1;
pub const TYPE_BLOCK: u32 = 2;
pub const TYPE_VSOCK: u32 = 19;
```

In our case the instance_id is a String provided by the user when the vm is configured. That is unique per device type. Having a key composed of device_type and this instance_id makes it easy for us to retrieve the device when the user sends a request for changing some parameters (changing the file path for a block device for example). So do you think the instance_id could be a String in your case also ? Or do you need it to be a global counter ?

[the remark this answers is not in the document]
liujing2: @serban300 Thanks for the example for device_type.  That is a good way to be light!
Umm, till now, the issue seems different with beginning :) It seems we feel a name string (e.g. DummyDevice::new(name)) can not ensure a unique per instance, so we extend it to a structure.
If we have structure {DeviceType , instance_id_string}, I guess we go back to the problem again? :-p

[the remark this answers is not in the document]
serban300: Yes, true ...

Again an enum would offer the flexibility of chosing what you need depending on the usecase, but it would be heavier. It might also look hacky. I don't know. Do you think it would make sense ?

```
pub enum DeviceInstanceId {
    GlobalCounter(u32),
    Name(String),
}
```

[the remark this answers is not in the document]
liujing2: @serban300 If the GlobalCounter(idx) is offered by users, or allocated by system?
I guess, we have an allocator, specifying idx to each new instance. But this needs it to be a struct... Something like,
```
pub struct InstanceId {
    pub last_id: u32,
}
impl InstanceId {
    /// Called by system
    pub fn new(start: u32) -> Self {
        InstanceId {last_id: start},
    }
    /// Get a u32 for each instance info {device_type: enum DeviceType, instance_id: u32}
    pub fn allocate() -> u32 {
        self.last_id = self.last_id + 1;
        self.last_id - 1
    }
}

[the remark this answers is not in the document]
Comment (serban300) [reply src/device_manager.rs]: It should be allocated by the system in order to guarantee its uniqueness. An allocator struct would work.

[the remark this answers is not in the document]
Comment (andreeaflorescu) [reply src/device_manager.rs]: How do you guarantee the uniqueness? If you use the `rand` crate in Rust it pulls a lot of dependencies. If we have to use rand, I would rather have the unique ID specified by the user of the crate so we don't pull all those dependencies in this library.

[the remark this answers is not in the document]
Comment (liujing2) [reply src/device_manager.rs]: @andreeaflorescu Why do we need `rand` crate? I think an allocator struct above can help us allocate a unique number to each instance. Do I miss anything?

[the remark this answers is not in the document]
Comment (andreeaflorescu) [reply src/device_manager.rs]: @liujing2 I totally missed the example code from your previous comment. It looks like we don't need rand.

Some comments:
- we need to make sure the id doesn't overflow
- `allocate` might be called from different threads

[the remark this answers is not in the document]
Comment (sameo) [reply src/device_manager.rs]: A few comments/questions:

* The ID allocation API would belong to the `vm-allocator`. As @andreeaflorescu pointed out, it should handle concurrency. Maybe implement this API as wrapper on top of the `Atomic` APIs?

* Are we agreeing on the fact that the key to the `devices` hash should be a structure, and not a string:

```Rust
pub struct DeviceId {
     /// An allocated ID from vm-allocator.
     /// Should we have a specific type for it?
     id: u32

     /// Device friendly name
     name: String
}
```

* Or should we instead have `vm-allocator` provide a naming API that would generate unique strings (given a prefix), and use that as a unique name for the device?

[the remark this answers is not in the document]
andreeaflorescu: @sameo I am a bit behind we the design of vm-device and vm-allocator. Do we want vm-device to take a dependency on vm-allocator?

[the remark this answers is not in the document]
liujing2: > Some comments:
> 
> * we need to make sure the id doesn't overflow
> * `allocate` might be called from different threads

@andreeaflorescu Yep, a u32 type for id can be enough. 
@sameo I agreed to make the id allocator in vm-allocator crate.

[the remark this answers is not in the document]
liujing2: > A few comments/questions:
> 
> * The ID allocation API would belong to the `vm-allocator`. As @andreeaflorescu pointed out, it should handle concurrency. Maybe implement this API as wrapper on top of the `Atomic` APIs?
> * Are we agreeing on the fact that the key to the `devices` hash should be a structure, and not a string:
> 
> ```rust
> pub struct DeviceId {
>      /// An allocated ID from vm-allocator.
>      /// Should we have a specific type for it?
>      id: u32
> 
>      /// Device friendly name
>      name: String
> }
> ```

According to above discussion with @serban300 's , they might prefer using enum instead of String as the name.

> * Or should we instead have `vm-allocator` provide a naming API that would generate unique strings (given a prefix), and use that as a unique name for the device?

I prefer to this idea. An API for allocating instance id (u32).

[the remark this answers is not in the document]
andreeaflorescu: If vm-device doesn't already have a dependency on vm-allocator I wouldn't add one just for this reason. We can have a DeviceId that needs to be passed when initializing a new device and let the caller decide how to initialize it. One might choose to use vm-allocator for that, another one can choose to set the device id using a different mechanism.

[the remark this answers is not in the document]
sameo: > According to above discussion with @serban300 's , they might prefer using enum instead of String as the name.

That would give names like e.g. `virtio-net-afb3467`, once the enum is stringified. Enums can be very generic as opposed to something more grepable based on a friendly, custom name provided by the device implementation as e.g. a prefix.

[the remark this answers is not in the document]
sameo: > @sameo I am a bit behind we the design of vm-device and vm-allocator. Do we want vm-device to take a dependency on vm-allocator?

It  does have a dependency on it already, this PR comes with a vm-allocator implementation for allocating and tracking pin based irqs and IO address spaces.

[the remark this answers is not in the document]
liujing2: @sameo, do you plan to submit vm-allocator crate adding this new API? Or I can update assuming there is one.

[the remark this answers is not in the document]
sameo: @liujing2 Please continue updating the vm-allocator code that's part of this PR, we will base the crate on it.

liujing2:
@serban300 @andreeaflorescu  @sameo  Two more patches for the naming issue was updated, Thanks.

src/device_manager.rs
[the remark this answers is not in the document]
liujing2: Comments addressed.

vm-allocator/src/system.rs
```diff
@@ -81,6 +84,16 @@ impl SystemAllocator {
...
+    /// Reserves the next available system device instance id number.
+    pub fn allocate_instance_id(&mut self) -> Option<u32> {
+        if let Some(id) = self.next_instance_id.checked_add(1) {
+            self.next_instance_id = id;
+            Some(id - 1)
+        } else {
+            None
+        }
+    }
+
```
sameo: Ok, so this is exactly the same code as the irq one. I think instead we need to build a generic `IdAllocator` structure and implement `allocate` and `free` methods for it. `allocate` could simply do the `checked_add()` that you do here but eventually we could decide to do something smarter (and use freed IDs) but that would take more complex data structures.

Both the above system irq and id allocation would call into the `IdAllocator.allocate()`.

sameo: Also, we need to make the system allocator thread safe. I think we should have one mutex per system resource, keep the address and id allocators lock less and take the locks at the `system.rs` level.

src/device.rs
```diff
@@ -58,10 +57,27 @@ impl IoResource {
...
+    Pci,
+    /// Virtio device type.
+    Virtio(u32),
+    /// Legacy serial device type.
+    Serial,
+    /// Legacy i8042 device type.
+    I8042,
+    /// RTC device type.
+    RTC,
+}
```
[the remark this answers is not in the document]
liujing2: Yep, I am not that fan, and also felt this kind of enum can be complex and confused for some types of devices, like virtio, virtio mmio, virtio pci and pure pci... How about @serban300 ?

src/device.rs
```diff
@@ -2,15 +2,14 @@
...
 use std::sync::{Arc, Mutex};
 use vm_memory::{GuestAddress, GuestUsize};
 
 /// Trait for devices with basic functions.
 #[allow(unused_variables)]
 pub trait Device: Send {
-    /// Get the device name.
-    fn name(&self) -> String;
+    /// Get the device type.
+    fn device_type(&self) -> DeviceType;
```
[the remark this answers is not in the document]
liujing2: @sameo  `type` is a key word in Rust, so I add a prefix rather than using escape everywhere.

vm-allocator/src/system.rs
liujing2: @sameo 
@sameo 
> Also, we need to make the system allocator thread safe. I think we should have one mutex per system resource, keep the address and id allocators lock less and take the locks at the `system.rs` level.

Which codes will use system allocator except vm-device? Once device_manager wrap it with Arc Mutex, it can protect it?

liujing2: > Ok, so this is exactly the same code as the irq one. I think instead we need to build a generic `IdAllocator` structure and implement `allocate` and `free` methods for it. `allocate` could simply do the `checked_add()` that you do here but eventually we could decide to do something smarter (and use freed IDs) but that would take more complex data structures.
> 
> Both the above system irq and id allocation would call into the `IdAllocator.allocate()`.

I agree with this.  There needs free for those resource.

sameo: > Which codes will use system allocator except vm-device? 

There may be other parts of a VMM requesting for e.g. some address space range (e.g. VFIO).

> Once device_manager wrap it with Arc Mutex, it can protect it?

You want the top level API to be Mutex protected, so that multi threaded callers don't have to all implement a Mutex wrapper on top of it. In most cases callers will want thread safe code, but if they really need lock less API, they can use the lower level ones (`AddressAllocator`, etc).

sameo: > I agree with this. There needs free for those resource.

Indeed.

src/device_manager.rs
```diff
@@ -87,30 +89,42 @@ impl DeviceManager {
...
-    fn allocate_resources(&mut self, resources: &mut Vec<IoResource>) -> Result<()> {
+    // Allocate IO, IRQ and instance id resources.
+    // Return a Result wrapper of instance id.
+    fn allocate_resources(&mut self, resources: &mut Vec<IoResource>) -> Result<(u32)> {
         let mut alloc_idx = 0;
+        let id = self
+            .resource
+            .lock()
+            .expect("failed to acquire lock.")
+            .allocate_instance_id();
```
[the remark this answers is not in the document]
liujing2: It seems we didn't pass any error information from vm-allocator APIs before?

liujing2:
@sameo  All comments are addressed.

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn free_instance_id(&mut self, id: u32) {
+        self.instance_id.free(id);
+    }
+
+    /// Reserves a section of `size` bytes of IO address space.
+    pub fn allocate_io_addresses(
+        &mut self,
+        address: GuestAddress,
+        size: GuestUsize,
+    ) -> Option<GuestAddress> {
```
sameo: Part of my previous comments that may not have been clear enough is that we want the APIs to return `Result` with the appropriate error instead of `Option`. So both `system.rs` and `address.rs` should return a `Result`.

sameo: Also, another part of the comments were that our `system.rs` API should be thread safe, and that we should associate a Mutex with each individual allocator that's part of `SystemAllocator`. Have you addressed that comment?

liujing2: @sameo  Let me add the protection in system.rs. Updated.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+            .set_resources(resources, irq);
+
+        let descriptor = self.device_descriptor(id, name, dev, parent_bus, resources.to_vec(), irq);
+
+        // Insert bus/device to DeviceManager with parent bus
+        self.insert(descriptor)
+    }
+
+    /// Unregister a device from `DeviceManager`.
+    pub fn unregister_device(&mut self, instance_id: u32) -> Result<()> {
```
sameo: How does the caller (The VMM) know which id to pass to `unregister_device` ? Shouldn't we return the device allocated id from `register_device`?

liujing2: @sameo, It seems this function would be called e.g. unhotplug. Otherwise, we won't do that right? If so, there needs a place to store and an API which might be supported later to list the device tree. So people would know id.

sameo: @liujing2 I don't think we should assume any particular use case here. I should be able to register/unregister a device from the device manager at will, for example for testing. Since the handle for registering a device is its newly allocated id, we should return it from the register call and let the caller store id for calling unregister with it.

sameo:
@serban300 @jiangliu @zachreizner I think this one is ready to be reviewed, could you PTAL?

jiangliu:
Great, will review it soon:)

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,208 @@
...
+pub struct SystemAllocator {
+    io_address_space: Arc<Mutex<Option<AddressAllocator>>>,
+    mmio_address_space: Arc<Mutex<AddressAllocator>>,
+    irq: Arc<Mutex<IdAllocator>>,
+    instance_id: Arc<Mutex<IdAllocator>>,
+}
+
+impl SystemAllocator {
+    /// Creates a new `SystemAllocator` for managing addresses and irq numvers.
+    /// Can return `None` if `base` + `size` overflows a u64 or if alignment isn't a power
```
jiangliu: the comments is out of date:)

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,208 @@
...
+/// ```
+pub struct SystemAllocator {
+    io_address_space: Arc<Mutex<Option<AddressAllocator>>>,
+    mmio_address_space: Arc<Mutex<AddressAllocator>>,
+    irq: Arc<Mutex<IdAllocator>>,
+    instance_id: Arc<Mutex<IdAllocator>>,
+}
+
+impl SystemAllocator {
+    /// Creates a new `SystemAllocator` for managing addresses and irq numvers.
```
jiangliu: numvers -> numbers?

src/device.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn name(&self) -> String;
+    /// Read from the guest physical address `addr` to `data`.
+    fn read(&mut self, addr: GuestAddress, data: &mut [u8], io_type: IoType);
+    /// Write `data` to the guest physical address `addr`.
+    fn write(&mut self, addr: GuestAddress, data: &[u8], io_type: IoType);
+    /// Set the allocated resource to device.
+    ///
+    /// This will be called by DeviceManager::register_device() to set
+    /// the allocated resource from the vm_allocator back to device.
+    fn set_resources(&mut self, res: &[IoResource], irq: Option<IrqResource>);
```
jiangliu: I think this is really to initialize the device instance,  so how about renaming this to something like initialize()/setup() and returning a Result?

src/device.rs
```diff
@@ -0,0 +1,96 @@
...
+}
```
jiangliu: Do we need destroy() here? Otherwise the device will be static by design.

src/device.rs
```diff
@@ -0,0 +1,96 @@
...
+    Mmio,
+    /// Non-exit physically backed mmap IO
+    PhysicalMmio,
+}
+
+/// Device resource information.
+#[derive(Debug, Copy, Clone)]
+pub struct IoResource {
+    /// Resource address.
+    pub addr: Option<GuestAddress>,
```
jiangliu: Why do we need Option<> here?

src/device.rs
jiangliu: And feels it would be easier to pass res and irq by value instead of by reference.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+    end: u32,
+    used_map: Vec<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new `IdAllocator` for managing u32 usage.
+    /// * `start` - The starting number to manage.
+    /// * `end` - The ending number to manage.
+    /// * `used_map` - The used numbers ordered from lowest to highest.
+    pub fn new(start: u32, end: u32) -> Option<Self> {
```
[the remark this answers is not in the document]
liujing2: It depends on what AddressAllocator and IdAllocator return. Not always Self.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    /// Allocates a number from the managed region. Returns `Ok(allocated_id)`
+    /// when successful, or Error indicates the failure reason.
+    pub fn allocate(&mut self, number: Option<u32>) -> Result<u32> {
+        let new = match number {
+            // Specified number to be allocated.
+            Some(num) => {
+                if num < self.start {
```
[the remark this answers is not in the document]
liujing2: Oh, yes. Will change it.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn first_usable_number(&self) -> Option<u32> {
+        if self.used_map.is_empty() {
+            return Some(self.start);
+        }
+
+        let mut previous = self.start;
+
+        for iter in self.used_map.iter() {
+            // We know the subtraction could not be invalid.
+            if (*iter - previous) > 0 {
```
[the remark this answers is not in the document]
liujing2: Right. Will change it.

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+///
+/// ```
+/// # use vm_allocator::AddressAllocator;
+/// # use vm_memory::{Address, GuestAddress, GuestUsize};
+///   AddressAllocator::new(GuestAddress(0x1000), 0x10000, Some(0x100)).map(|mut pool| {
+///       assert_eq!(pool.allocate(None, 0x110), Some(GuestAddress(0x10e00)));
+///       assert_eq!(pool.allocate(None, 0x100), Some(GuestAddress(0x10c00)));
+///   });
+/// ```
+#[derive(Debug, Eq, PartialEq)]
```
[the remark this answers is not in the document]
liujing2: Since address.rs has some tests needing the properties.

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+        let mut allocator = AddressAllocator {
+            base,
+            end,
+            alignment,
+            ranges: BTreeMap::new(),
+        };
+
+        // Insert the last address as a zero size range.
+        // This is our end of address space marker.
+        allocator.ranges.insert(base.checked_add(size)?, 0);
```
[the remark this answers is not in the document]
liujing2: If we want to keep the `end` being inclusive, I see another place needs to call checked_sub(1) once we define `end` like that :)
```
        let mut allocator = AddressAllocator {
            base,
            end.checked_sub(1),
            alignment,
            ranges: BTreeMap::new(),
        };
```

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+    ) -> Result<GuestAddress> {
+        let aligned_address = self.align_address(req_address);
+
+        // The requested address should be aligned.
+        if aligned_address != req_address {
+            return Err(Error::UnalignedAddress);
+        }
+
+        // The aligned address should be within the address space range.
+        if aligned_address >= self.end || aligned_address <= self.base {
```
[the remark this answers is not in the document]
liujing2: Right! Will change it.

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    fn align_address(&self, address: GuestAddress) -> GuestAddress {
+        let align_adjust = if address.raw_value() % self.alignment != 0 {
+            self.alignment - (address.raw_value() % self.alignment)
+        } else {
+            0
+        };
+
+        address.unchecked_add(align_adjust)
```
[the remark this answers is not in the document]
liujing2: Yep, will change the function return Option<> and use checked_add instead.

vm-allocator/src/address.rs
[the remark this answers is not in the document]
jiangliu: The key point is to remove the error handle path in allocator.ranges.insert(base.checked_add(size)?, 0);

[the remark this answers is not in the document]
liujing2: Yep, got the point. I mean we need minus 1 again to store the real end number. But we can directly use unchecked_sub(1) since size>=0.
```
        let bound = base.checked_add(size)?;
        let end = bound.unchecked_sub(1); 
        let mut allocator = AddressAllocator {
            base,
            end,
            alignment,
            ranges: BTreeMap::new(),
        };
         allocator.ranges.insert(bound, 0);
```

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+            // we return the start of this range minus the requested size.
+            // As each new range is allocated at the end of the available address space,
+            // we will tend to always allocate new ranges there as well. In other words,
+            // ranges accumulate at the end of the address space.
+            if address
+                .unchecked_sub(self.align_address(prev_end_address).raw_value())
+                .raw_value()
+                >= req_size
+            {
+                return Ok(self.align_address(address.unchecked_sub(req_size + self.alignment)));
```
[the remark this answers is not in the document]
liujing2: You'r right. Also this would have hole when req_size is aligned. What about this one?
```
       let req_align = self.align_address(GuestAddress(req_size))?;
       let addr = self.align_address(address.unchecked_sub(req_align.raw_value()))?;
```

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+pub type Result<T> = result::Result<T, Error>;
+
+/// Manages allocating unsigned 32-bit number usage.
+/// Use `IdAllocator` whenever an unsigned 32-bit number needs to be allocated to different users.
+///
+/// # Examples
+///
+/// ```
+/// # use vm_allocator::IdAllocator;
+///   IdAllocator::new(1).map(|mut p| {
```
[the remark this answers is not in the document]
andreeaflorescu: The tests should have failed if the example is out of date. Why are the tests not failing for this case?

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        self.ranges.insert(new_addr, size);
+
+        Ok(new_addr)
+    }
+
+    /// Free an already allocated address range.
+    /// We can only free a range if it matches exactly an already allocated range.
+    pub fn free(&mut self, address: GuestAddress, size: GuestUsize) {
+        if let Some(&range_size) = self.ranges.get(&address) {
```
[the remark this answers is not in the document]
liujing2: Right. Thanks!

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+    AddressAllocate(AllocatorError),
+}
+
+/// Simplify the `Result` type.
+pub type Result<T> = result::Result<T, Error>;
+
+/// System device manager serving for all devices management and VM exit handling.
+pub struct DeviceManager {
+    /// System allocator reference.
+    resource: Arc<Mutex<SystemAllocator>>,
```
jiangliu: All fields in SystemAllocator are already protected by Arc<Mutex<>>, why another level of Arc<Mutex<>>?
Seems we need to specify who takes the responsibility to protect DeviceManager instance from concurrent access. Suggestion the owner of the DeviceManager object takes the responsibility to protect from concurrent access. If the vmm, such as firecracker,  supports static device manager, there's need for Mutex at all. On the other hand, VMM like qemu may protect the DeviceManager object withj Arc<Mutex<>>.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// Devices information mapped by instance id.
+    devices: HashMap<u32, DeviceDescriptor>,
+    /// Range mapping for VM exit mmio operations.
+    mmio_bus: BTreeMap<Range, Arc<Mutex<dyn Device>>>,
```
jiangliu: I have strong concern about the design decision to protect Device object with Arc<Mutex<>> at the DeviceManager layer. That means all accesses to the Device object will be serialized by the device manager, and virtio multi-queues features will be effectively disabled by this design. Actually we should let device backend to decide how to protect its fields.
So mmio_bus: BTreeMap<Range, Arc<Mutex<dyn Device>>> should be
mmio_bus: BTreeMap<Range, Arc<dyn Device>> instead.
And a DummyDevice may be implemented as :
struct DummyDevice {
     ro_field: u32,
    rw_field: RwLock<u32>,
    rw_field2: Mutex<u32>,
};

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+            resource,
+            devices: HashMap::new(),
+            mmio_bus: BTreeMap::new(),
+            pio_bus: BTreeMap::new(),
+        }
+    }
+
+    fn insert(&mut self, dev: DeviceDescriptor) -> Result<()> {
+        // Insert if the key is non-present, else report error.
+        if self.devices.get(&(dev.instance_id)).is_some() {
```
jiangliu: BtreeMap::contains_key() is designed for this:)

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+                            .allocate_mmio_addresses(res.addr, res.size)
+                            .map_err(Error::AddressAllocate)?,
+                    )
+                }
+            }
+        }
+        Ok(id)
+    }
+
+    fn free_resources(&mut self, resources: &[IoResource], id: u32) {
```
jiangliu: let res = self
                    .resource
                    .lock()
                    .expect("failed to acquire lock");
so we don't to lock/unlock for every resource entry.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+                    .free_mmio_addresses(res.addr.unwrap(), res.size),
+            }
+        }
+        self.resource
+            .lock()
+            .expect("failed to acquire lock")
+            .free_instance_id(id);
+    }
+
+    fn register_resources(
```
jiangliu: unregister_resoures() may be needed too.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+                    res.addr = Some(
+                        self.resource
+                            .lock()
+                            .expect("failed to acquire lock")
+                            .allocate_io_addresses(res.addr.unwrap(), res.size)
+                            .map_err(Error::AddressAllocate)?,
+                    );
+                }
+                IoType::PhysicalMmio | IoType::Mmio => {
+                    res.addr = Some(
```
jiangliu: We need a flag to mark whether the resource is dynamically allocated from the resource pool or statically reserved. So we could do the right thing when frees the resource.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+            .lock()
+            .expect("failed to acquire lock.")
+            .allocate_instance_id()
+            .map_err(Error::InstanceIdAllocate)?;
+
+        for res in resources.iter_mut() {
+            match res.res_type {
+                IoType::Pio => {
+                    if res.addr.is_none() {
+                        return Err(Error::NonePIORequest);
```
jiangliu: Allocated resources, such as the id, should be released on errors.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+    ) -> Result<()> {
+        for res in resources.iter() {
+            match res.res_type {
+                IoType::Pio => {
+                    if self
+                        .pio_bus
+                        .insert(Range(res.addr.unwrap(), res.size), dev.clone())
+                        .is_some()
+                    {
+                        return Err(Error::Overlap);
```
jiangliu: Error recover is needed too, otherwise the behavior may be undefined.

src/device_manager.rs
```diff
@@ -0,0 +1,431 @@
...
+        dev: Arc<Mutex<dyn Device>>,
+        parent_bus: Option<Arc<Mutex<dyn Device>>>,
+        resources: &mut Vec<IoResource>,
+        interrupt: Option<IrqResource>,
+    ) -> Result<()> {
+        // Reserve resources
+        let id = self.allocate_resources(resources)?;
+
+        // Register device resources
+        if let Err(Error::Overlap) = self.register_resources(dev.clone(), resources) {
```
jiangliu: Please do error recover too

jiangliu:
Seems more thoughts are needed about the DeviceManager abstraction. Especially the way to support both simple VMM like firecracker and full functionally vmm like qemu, they have different requirements.

vm-allocator/src/id.rs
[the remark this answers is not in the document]
liujing2: Will update the fix soon. I think it is because vm-allocator/ is behind vm-device/ and seems not covered when execute commands under vm-device/

src/device_manager.rs
liujing2: Yes, I have a new patch removing the protection here, since we want SystemAllocator to be thread safe of each members, and also, vmm would have a static DeviceManager being protected.

src/device.rs
liujing2: Because we needs to separate a specified address request and a non-specific address request. The vm-allocator::address::allocate shows the details.

jiangliu: Thanks for explanation. Then a helper function is welcomed which assumes that  the `addr` field contains a valid address and returns it.  So callers don't need to deal with the Option<> wrapper.

src/device_manager.rs
liujing2: Got it , thanks!

src/device_manager.rs
liujing2: Once we protect DeviceManager in VMM, and keep SystemAllocator members thread safe, the resources are protected. Do I understand correctly for your comments? (The codes with the last updated patch)

src/device_manager.rs
liujing2: So let's create a explicit function for the resources freeing, instead of inside unregister_device().

src/device_manager.rs
liujing2: Do you mean a specified address request and non-specified one? If so, both of them are being allocated and marked there. And both of them can be freed when unregister_device.

src/device_manager.rs
liujing2: Oh, right!

src/device_manager.rs
liujing2: Do you mean `res.addr.unwrap()`? Or anything else? Sorry I didn't quite get it.

src/device.rs
liujing2: Do you mean unregister_device? That's kind of destroy to free resources and unregister things.

src/device.rs
liujing2: Sounds good! Will add that in update.

src/device_manager.rs
liujing2: Ok, got it. Will return the instance id for vmm managing in update.

src/device_manager.rs
liujing2: I guess you mean some operations here to let register succeed?

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright 2018 The Chromium OS Authors. All rights reserved.
+// Copyright © 2019 Intel Corporation
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE-BSD-3-Clause file.
+//
+// SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
```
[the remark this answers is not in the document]
sameo: Yes.

src/device.rs
sameo: > I think this is really to initialize the device instance, so how about renaming this to something like initialize()/setup() and returning a Result?

It's more of a notification back to the device than an initialization. The resources have been allocated by the system allocator and it's telling the device about it. It's up to the device to decide what to do with that information.

src/device_manager.rs
sameo: > virtio multi-queues features will be effectively disabled by this design

Could you please elaborate on that?

sameo:
> Seems more thoughts are needed about the DeviceManager abstraction. Especially the way to support both simple VMM like firecracker and full functionally vmm like qemu

The current design is driven by the assumption that the VMM will be the central resource allocation point, which is true for firmware-less and simple VMMs like e.g. Firecracker or to some extend crosvm. It's also a little static at the moment as it does not yet allow for device resources reprogramming (e.g. PCI).
However, I think this is a good starting point and it conveys the design of what both Firecracker and crosvm do. From there, we can extend the device manager API to support more complex use cases.

> they have different requirements.

Maybe we could try to enumerate those requirements and see if the current PR could eventually also handle them. Besides device driven resources re-allocation (aka reprogramming), what else do you think we should take into account?

src/device_manager.rs
sameo: > We need a flag to mark whether the resource is dynamically allocated from the resource pool or statically reserved.

What do you mean by statically reserved? Physical RAM for example?

src/device_manager.rs
liujing2: @sameo I guess it means multiple queues might notify vmm that some buffers should be handled at the same time, with different threads?

src/device.rs
jiangliu: I'm thinking about how to support device hot plug/unplug.
On boot:
1) create a device instance.
2) register the device instance to the device manager.
3) allocate resources for all registered devices.
4) start all registered devices

On hot-add:
1) create device instance
2) register device instance to the device manager
2.1) allocate resources for the device
2.2) start the device

On hot-removal:
1) look up the device
2) stop the device
3) free resources of the device
4) remove the device from the device manager
5) destroy the device

Seems there  are still some missing interfaces to support above flow. 
So I suggested to set_resources() as start(), and add missing stop().

src/device_manager.rs
jiangliu: Yes. The device may be accessed by different vcpus concurrently, but current design of the device manager enforces serial accesses due to the Arc<Mutex<dyn Device>>.

src/device_manager.rs
jiangliu: I means just acquire the lock once at entries, instead of acquire/release for each accessed fields.

src/device_manager.rs
jiangliu: > > Seems more thoughts are needed about the DeviceManager abstraction. Especially the way to support both simple VMM like firecracker and full functionally vmm like qemu
> 
> The current design is driven by the assumption that the VMM will be the central resource allocation point, which is true for firmware-less and simple VMMs like e.g. Firecracker or to some extend crosvm. It's also a little static at the moment as it does not yet allow for device resources reprogramming (e.g. PCI).
> However, I think this is a good starting point and it conveys the design of what both Firecracker and crosvm do. From there, we can extend the device manager API to support more complex use cases.
> 
> > they have different requirements.
> 
> Maybe we could try to enumerate those requirements and see if the current PR could eventually also handle them. Besides device driven resources re-allocation (aka reprogramming), what else do you think we should take into account?

Sure, we focus on simple static alike VMM too:)

jiangliu: > > We need a flag to mark whether the resource is dynamically allocated from the resource pool or statically reserved.
> 
> What do you mean by statically reserved? Physical RAM for example?

Actually I hope to make free_resources() undo all changes made by allocate_resources().
For an IoResource entry 'res', allocate_resoures() will set res.addr to Some(addr) on success when called with res.addr as None. So I hope res.addr will be reset to None after calling free_resources().

src/device_manager.rs
liujing2: There might have a problem: we should assume and limit the device instance to be only one pointer. Because when we do `dev.set_resources()` which needs mutable self, `Arc<> provides related APIs like `get_mut()` and `make_mut()` both assuming there's one pointer, or else it would clone the inner value or panic. So we should also limit the VMM not clone pointer before register_device. Use get_mut() to use panic to protect unique inner value.
```
    pub fn register_device(
        &mut self,
        dev: Arc<dyn Device>,
         ...
    ) -> Result<()> {
             ...
        dev.get_mut().set_resources(resources);
             ...
    }
```

src/device_manager.rs
liujing2: > > > We need a flag to mark whether the resource is dynamically allocated from the resource pool or statically reserved.
> > 
> > 
> > What do you mean by statically reserved? Physical RAM for example?
> 
> Actually I hope to make free_resources() undo all changes made by allocate_resources().
> For an IoResource entry 'res', allocate_resoures() will set res.addr to Some(addr) on success when called with res.addr as None. So I hope res.addr will be reset to None after calling free_resources().

Thanks for the detailed explanation.  Referring to [above comment](https://github.com/rust-vmm/vm-device/pull/3#discussion_r300038235), I' thinking if it is really need to set original resource status back to device instance when unhotplug device. Because device instances are managed by DeviceManager and all operations are not able to be executed after unregister_device(). And next time, when re-hotplug this device, it will re-tell what kinds of resouces it needs.

src/device.rs
liujing2: > And feels it would be easier to pass res and irq by value instead of by reference.

Could I ask why pass res as Vec<> instead of &[] would be easier/simple?  BTW, this function is called inside register_device(..., resources: &mut Vec<IoResource>)

liujing2:
@sameo @jiangliu @andreeaflorescu 
PR is updated. Some comments are not resolved and keep them as it is for now. Would update more after more discussion.

src/device_manager.rs
```diff
@@ -0,0 +1,461 @@
...
+                self.free_io_resources(&resources[0..idx - 1]);
+                self.free_id_resource(id);
+                return Err(Error::IoResourceAllocate(idx, e));
+            }
+        }
+
+        // Register device resources
+        let register_len = self.register_resources(dev.clone(), resources);
+        // Unregister and free resources once failed.
+        if register_len < resources.len() && register_len > 0 {
```
jiangliu: We should release io resources and id even if register_len is 0.

liujing2: Yep, will change it.

src/device_manager.rs
liujing2: As discussed with @sameo , the key point here is that Arc<> limits the pointer numbers once we need direct the inner value to be mutable referece. But with Mutex<>, MutexGuard help impl DerefMut trait so things become easier. 

I think we can set all functions of `Device` trait to be `&self` instead of `&mut self`, so that DeviceManager don't need care mutable inner device and let device structure decide which members should be `Mutex` protection to solve this problem. Let's update the codes to discuss more.

jiangliu: Yes, it means to design the device trait with interior mutability.

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+    IdAllocate(crate::id::Error),
+}
+
+impl Display for Error {
+    // This trait requires `fmt` with this exact signature.
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::Error::*;
+
+        match self {
+            NoneAddress => write!(f, "The address being allocated is null"),
```
andreeaflorescu: `null` is improperly said since null doesn't really exist in its pure form in Rust. Also, the emphasize should be on the operation that failed (i.e. the allocation) and not on the object being used. I would rephrase this to: "Cannot allocate address. The I/O address space is undefined." This would require splitting the errors per address space as I suggested in my previous comment.

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Address is none.
+    NoneAddress,
+    /// Address allocation failed.
+    AddressAllocate(crate::address::Error),
+    /// Id allocation failed.
+    IdAllocate(crate::id::Error),
+}
+
+impl Display for Error {
+    // This trait requires `fmt` with this exact signature.
```
andreeaflorescu: This comment is not necessary.

vm-allocator/src/address.rs
```diff
@@ -0,0 +1,381 @@
...
+    end: GuestAddress,
+    alignment: GuestUsize,
+    ranges: BTreeMap<GuestAddress, GuestUsize>,
+}
+
+impl AddressAllocator {
+    /// Creates a new `AddressAllocator` for managing a range of addresses.
+    /// Can return `None` if `base` + `size` overflows a u64 or if alignment isn't a power
+    /// of two.
+    ///
```
andreeaflorescu: The following lines should be placed in an `#Arguments` section.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Debug)]
+pub struct IdAllocator {
+    start: u32,
+    end: u32,
+    used_map: Vec<u32>,
+}
+
+impl IdAllocator {
+    /// Creates a new `IdAllocator` for managing u32 usage.
+    /// * `start` - The starting number to manage.
```
andreeaflorescu: The explanation for the fields should be place in the structure before each field. `used_map` is not an argument for the new function.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+///   IdAllocator::new(1, std::u32::MAX).map(|mut p| {
+///       assert_eq!(p.allocate(Some(1)).unwrap(), 1);
+///       assert_eq!(p.allocate(Some(3)).unwrap(), 3);
+/// });
+/// ```
+#[derive(Debug)]
+pub struct IdAllocator {
+    start: u32,
+    end: u32,
+    used_map: Vec<u32>,
```
andreeaflorescu: I find `used_map` to be a misleading name because you actually use a vector.

vm-allocator/src/system.rs
andreeaflorescu: I am not sure we can use this code as is with arc and mutex on each resource. Maybe we should provide a trait here instead? Allocating these resources is on the critical path (i.e. it adds to the boot time) for Firecracker.

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+///    assert_eq!(allocator.allocate_irq(Some(7)).unwrap(), 7);
+///    assert_eq!(allocator.allocate_mmio_addresses(None, 0x1000).unwrap(), GuestAddress(0x1ffff000));
+///
+/// ```
+#[derive(Clone)]
+pub struct SystemAllocator {
+    io_address_space: Arc<Mutex<Option<AddressAllocator>>>,
+    mmio_address_space: Arc<Mutex<AddressAllocator>>,
+    irq: Arc<Mutex<IdAllocator>>,
+    instance_id: Arc<Mutex<IdAllocator>>,
```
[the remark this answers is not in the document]
liujing2: It is a unique number for each device instance, allocated by system allocator from a start number. As [discussed  before](https://github.com/rust-vmm/vm-device/pull/3#discussion_r291114147) :)

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+use std::result;
+use std::sync::{Arc, Mutex};
+
+/// Errors associated with system resources allocation.
+#[derive(Debug)]
+pub enum Error {
+    /// Address is none.
+    NoneAddress,
+    /// Address allocation failed.
+    AddressAllocate(crate::address::Error),
```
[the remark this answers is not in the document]
liujing2: This code is updated with later patch after IRQ is supported so the last view of vm-device is like this,
```
    IoResourceAllocate(usize, AllocatorError),
    /// IRQ allocated failed.
    IrqAllocate(AllocatorError),
    /// Instance id allocation failed.
    InstanceIdAllocate(AllocatorError),
}
```

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Creates a new `SystemAllocator` for managing addresses and irq numbers.
+    ///
+    /// * `io_base` - The starting address of IO memory.
+    /// * `io_size` - The size of IO memory.
+    /// * `mmio_base` - The starting address of MMIO memory.
+    /// * `mmio_size` - The size of MMIO memory.
+    /// * `first_irq` - The first irq number to give out.
+    /// * `last_irq` - The last irq number to give out.
+    /// * `first_instance_id` - The first device instance id to give out.
+    pub fn new(
```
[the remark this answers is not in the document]
liujing2: `first_instance_id` means the start number given to the first device instance. Then, system allocator would know where to allocate from.

Make sense of the setter function. But when if we want to allocate another type of resource in the future, the SystemAllocator would also add a new member so the constructor would also change, right?

vm-allocator/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "vm-allocator"
+version = "0.1.0"
+authors = ["The Chromium OS Authors"]
+edition = "2018"
+
+[dependencies]
+libc = "*"
```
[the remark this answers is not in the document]
liujing2: Thanks for the information! It really needs change to some specific version. I'll modify it in next update.

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+///           GuestAddress(0x10000000), 0x10000000,
+///           5, 24, 1).unwrap();
+///    assert_eq!(allocator.allocate_irq(None).unwrap(), 5);
+///    assert_eq!(allocator.allocate_irq(Some(7)).unwrap(), 7);
+///    assert_eq!(allocator.allocate_mmio_addresses(None, 0x1000).unwrap(), GuestAddress(0x1ffff000));
+///
+/// ```
+#[derive(Clone)]
+pub struct SystemAllocator {
+    io_address_space: Arc<Mutex<Option<AddressAllocator>>>,
```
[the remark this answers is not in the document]
liujing2: Because port IO address space would be None in some modern VMM, but MMIO address space should exist.
But you're right that the legacy irq resource also needs to be Option. @sameo How do you think?

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+                    return Err(Error::Overflow);
+                }
+                match self.used_map.iter().find(|&&x| x == num) {
+                    Some(_) => {
+                        return Err(Error::Duplicated);
+                    }
+                    None => num,
+                }
+            }
+            None => self.first_usable_number().ok_or(Error::Overflow)?,
```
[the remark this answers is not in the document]
liujing2: Right, `first_usable_number` returning None means Overflow, so do you mean change the following case error type?
```
            // Specified number to be allocated.
            Some(num) => {
                if num < self.start || num > self.end {
                    return Err(Error::Invalid);
                }
```

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// * `used_map` - The used numbers ordered from lowest to highest.
+    pub fn new(start: u32, end: u32) -> Option<Self> {
+        Some(IdAllocator {
+            start,
+            end,
+            used_map: Vec::new(),
+        })
+    }
+
+    fn first_usable_number(&self) -> Option<u32> {
```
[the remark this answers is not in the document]
liujing2: This function would only tell us that there is no usable number because of all are used, in other words, there's one error case only. Do we still need recover for such a function that we exactly know?

vm-allocator/src/system.rs
liujing2: Got it. So let's change the error as
UndefinedAddress => write!(f, "Cannot allocate address. The address being allocated is undefined"),
This kind of error would only happen during PIO allocation.

vm-allocator/src/system.rs
liujing2: Will remove it.

vm-allocator/src/id.rs
liujing2: Ok, got it! Thanks for the correct comment format reminder! :)

vm-allocator/src/id.rs
liujing2: Ummm, yep. What about `used` ?

vm-allocator/src/system.rs
liujing2: > I am not sure we can use this code as is with arc and mutex on each resource. Maybe we should provide a trait here instead? Allocating these resources is on the critical path (i.e. it adds to the boot time) for Firecracker.

@andreeaflorescu  
Sorry I didn't quite get the comments. Which part of codes does your comments for?

vm-allocator/src/system.rs
```diff
@@ -0,0 +1,207 @@
...
+use crate::id::IdAllocator;
+
+use libc::{sysconf, _SC_PAGESIZE};
+use std::fmt::{self, Display};
+use std::result;
+use std::sync::{Arc, Mutex};
+
+/// Errors associated with system resources allocation.
+#[derive(Debug)]
+pub enum Error {
```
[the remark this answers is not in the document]
liujing2: Thanks for guidance.  Let me improve all the error words in later update.

vm-allocator/src/id.rs
```diff
@@ -0,0 +1,119 @@
...
+        if previous <= self.end {
+            Some(previous)
+        } else {
+            None
+        }
+    }
+
+    /// Allocates a number from the managed region. Returns `Ok(allocated_id)`
+    /// when successful, or Error indicates the failure reason.
+    pub fn allocate(&mut self, number: Option<u32>) -> Result<u32> {
```
andreeaflorescu: Please split the comments into summary and description. I would also skip the part where you say that it returns Ok on success. The interesting part is the Error path. Rust [documentation guidelines](https://doc.rust-lang.org/1.30.0/book/2018-edition/ch14-02-publishing-to-crates-io.html#commonly-used-sections) specify that when a function can return error, you can describe when does the error occur by having `#Errors` section.

You could write something like:
```rust
\\\ Allocates a unique number from the managed region.
\\\ 
\\\ # Arguments
\\\
\\\ *`number` - If the `Option` is `Some`, the function reserves the specified number.
\\\  
\\\ # Errors
\\\
\\\ When `number` has a value, it returns an error when the number is out of the defined interval or the
\\\ number was previously reserved in a previous call of the function. When the number is not specified,
\\\ it returns an error when there are no more numbers available in the interval.
```

vm-allocator/src/id.rs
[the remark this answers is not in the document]
andreeaflorescu: It was more of a nit. If you return `Result` here instead of `Option` you don't need to do a match on the result of this function and just call `first_usable_number()?` which simplifies the code a bit.

andreeaflorescu:
I looked a bit more at the design of the DeviceManager as well as the SystemAllocator.

Right now the SystemAllocator is not very flexible because:
1. it has a static list of resources it manages, specified as fields of the SystemAllocator structure.
2. each resource allocator (io_address_space, mmio_address_space, ...) is held in an Arc<Mutex<..>> which makes sense for a multi-threaded VMM, but not really for Firecracker. This might not be a huge problem depending on how fast the operation on `Arc` are, but this still needs to be taken into account.

I tried to experiment a bit with creating a `ResourceAllocator` trait that could fix point 1, but it is complicating the code a bit more than is necessary. I changed the `SystemAllocator` to keep a hashmap of all `ResourceAllocator` objects as follows:

```rust
/// Trait for defining resources which can be allocated.
///
/// This trait needs to be implemented by all resources that are to be managed
/// by a `ResourceAllocator`. Example of resources are Guest Addresses, IRQ numbers and others.
pub trait Resource {}

/// Trait for defining a resource allocator.
///
/// Resources which should implement a resource allocator are IRQs, Virtual Memory, MMIO, and
/// other resources for which you have a space definition (like an interval of valid addresses)
/// and you can want to make sure that objects in the defined space are unique.
pub trait ResourceAllocator {
    fn allocate(&mut self) -> Result<Box<Resource>, Error>;
    fn free(&mut self) -> Result<(), Error>;
}

struct SystemAllocator {
    // The SystemAllocator resources are kept in a hashmap so that it is agnostic to the
    // number and type of resources it manages. For ease of use, the hashmap key should be
    // the name of the resource (for example "irq_allocator").
    resource_allocators: HashMap<String, Box<dyn ResourceAllocator>>,
}

impl SystemAllocator {
    pub fn new() -> Self {
        SystemAllocator {
            resource_allocators: HashMap::new()
        }
    }

    /// Add a resource allocator to be managed by `SystemAllocator`.
    // TODO: return result;
    pub fn add_allocator(&mut self, allocator_id: String, allocator: Box<ResourceAllocator>) {
        // TODO: key does not exist?
        self.resource_allocators.insert(allocator_id, allocator);
    }

    /// Allocate a new resource in the allocator specified by `allocator_id`.
    pub fn allocate(&mut self, allocator_id: String) -> Result<Box<Resource>, Error> {
        self.resource_allocators.get_mut(&allocator_id).unwrap().allocate(None)
    }
}
```

To test the new interface of the SystemAllocator, I did a dummy test with `IdAllocator`:

```rust
// We can't use the raw value `u32` because the `ResourceAllocator` trait can be used by multiple resources.
// For this reason we use `Resource` in `ResourceAllocator` and we implement `Resource` for `Id`.
pub struct Id(pub u32);
impl Resource for Id {}

pub struct IdAllocator {
    start: Id,
    end: Id,
    used_map: Vec<Id>,
}

impl ResourceAllocator for IdAllocator {
    fn allocate(&mut self) -> Result<Box<Resource>, Error> {
        self.used_map.push(Id(42));
        Ok(Box::new(Id(42)))
    }

    fn free(&mut self) -> Result<(), Error> {
        unimplemented!()
    }
}

```

This implementation is not thread safe, so the `SystemAllocator` would need to be placed in an `Arc` which might make the code inefficient for multi-threaded VMMs.

I honestly couldn't find a pattern that could work for multiple VMM implementations (Firecracker, CrosVM, Cloud Hypervisor) and I am starting to think that maybe it is a better idea to not make the `vm-device` depend on the implementation of `vm-allocator`. Instead of having `DeviceManager` call into `SystemAllocator`, we can pass a callback function to `register_device` and `unregister_device`. With this approach in Firecracker we would end up with a minimal function and for projects that need  complex resource management, the VMM can pass functions from the vm-allocator. This would require the vm-allocator to be a separate crate.

What do you all think?

P.S. I haven't done an in-depth review for the `DeviceManager` code.

andreeaflorescu:
Just discovered that there's a leftover travis configuration file. That should be deleted as well (.travis.yml)

vm-allocator/src/system.rs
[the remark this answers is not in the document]
andreeaflorescu: If it's a device id maybe we should name it `device_id`.

vm-allocator/src/system.rs
andreeaflorescu: The comment refers to the previous comment from @sameo. It is about having AddressAllocator and IdAllocator in a Mutex.

vm-allocator/src/system.rs
andreeaflorescu: `NoneAddress` is returned when `io_adress_space` is not initialized so I don't think that `UndefinedAddress` is an accurate name.

We can have a generic error like `UninitializedAllocator(&'static str)` which we can use for all allocators that are optional. When an `allocate_` function is called for an allocator that was not initialized, we can return `UninitializedAllocator("allocator_name")`.

```rust
pub enum Error {
    /// Allocator is not initialized.
    UninitializedAllocator(&'static str),
    /// MMIO allocation failed.
    AllocateMmio(crate::address::Error),
    /// I/O allocation failed.
    AllocateIO(crate::address::Error),
    /// IRQ allocation failed.
    AllocateIrq(crate::id::Error),
   /// device_id allocation failed.
    AllocateDeviceId(crate::id::Error),
}

impl Display for Error {
    // This trait requires `fmt` with this exact signature.
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use self::Error::*;

        match self {
            // When implementing display we can print the name of the uninitialized allocator so it makes it easier to debug. 
            UninitializedAllocator(allocator_name) => write!(f, "The allocator {} is not initialized.", allocator_name),
            ....  
        }
    }
}
```

Let's assume that both I/O Allocator and IRQ Allocator are optional. In this case both `allocate_io` and `allocate_irq` can fail with the same error `UninitializedAllocator`, but we can differentiate them because the error contains the name of the allocator that failed.

```rust
    /// Reserves a section of `size` bytes of IO address space.
    pub fn allocate_io_addresses(
        &mut self,
        address: GuestAddress,
        size: GuestUsize,
    ) -> Result<GuestAddress> {
        if let Some(io_allocator) = self
            .io_address_space
            .lock()
            .expect("failed to acquire lock")
            .as_mut()
        {
            ....
        } else {
            Err(Error::UninitializedAllocator("io_allocator"))
        }
    }

....

    pub fn allocate_irq(&mut self, irq: Option<u32>) -> Result<u32> {
         if let Some(irq_allocator) = self.irq
            .lock()
            .expect("failed to acquire lock") {
             ......
         } else {
            Err(Error::UninitializedAllocator("irq_allocator"))
         }
    }

```

liujing2:
@andreeaflorescu I like the idea of letting DeviceManager release reference of SystemAllocator! 

After thinking deep and experiments, I indeed found some key problems of design and let's discuss in details.

Firstly, let's say some necessary demands of design.
- To get raw value of each `Resource` request, it could be defined as,
```rust
pub trait Resource
{
    /// Type of the resource raw value, like u32 for Id.
    type V;
    /// Get the name describing the resource type.
    fn name(&self) -> String;
    fn raw_value(&self) -> Self::V;
}
```
- `ResourceAllocator` needs a parameter of `Resource` to allocate a specific one. With associated type V, `Resource` could not be `Box<Resource>` because V needs to be specified. So we use bounding.
```rust
pub trait ResourceAllocator {
    fn name(&self) -> String;
    fn allocate<R: Resource>(
        &mut self,
        res: Option<Box<R>>,
   ) -> Result<Box<R>, Error>;
    ...
}

impl ResourceAllocator for Id {
    ...
}
```
But the problem is, compiler couldn't know the detailed type of input `res.raw_value()`, but only expects type `<R as Resource>::V`, because input is trait. So, we actually could not analyse `res` in the allocate() function to do anything. Also, this makes things dangerous because no limited input type. (This has workaround to use name() to check Resource and ResourceAllocator)

Meanwhile, considering SystemAllocator::allocate() which returns `Result<Box<Resource<V=?>>>`, we don't know the right value for each allocator. I don't think this has workaround :(

========
Since the key is we want to keep abstraction in high level but compiler could not know details, I tried to split all resources into two: `AddrResourceAllocator` for PIO and MMIO and `UIntResourceAllocator` for irq and device_id. So far, I didn't find any other resources needed by system allocator.

```rust
pub trait UIntResourceAllocator {
    /// Unsigned integer resource allocation.
    fn allocate(
        &mut self,
        res: Option<Box<Resource<V = u32>>>,
   ) -> Result<Box<Resource<V = u32>>, Error>;
}

pub trait AddrResourceAllocator {
    /// Address resource allocation.
    fn allocate(
        &mut self,
        res: Option<Box<Resource<V = GuestAddress>>>,
    ) -> Result<Box<Resource<V = GuestAddress>>, Error>;
}

impl UIntResourceAllocator for IdAllocator {
    fn allocate(&mut self, resource: Option<Box<Resource<V=u32>>>) -> Result<Box<Resource<V=u32>>> {
        // resource is specified, resource.raw_value() is u32
    }
```
This requires SystemAllocator to have two members and two functions for each. 
```rust
pub struct SystemAllocator {
    addr_resource_alloc: HashMap<String, Box<AddrResourceAllocator>>,
    uint_resource_alloc: HashMap<String, Box<UIntResourceAllocator>>,
}

pub fn add_uint_allocator(&mut self, allocator_name: String, allocator: Box<UIntResourceAllocator>) -> Result<()> {...}

pub fn allocate_uint(&mut self, _allocator_id: String) -> Result<Box<Resource<V=u32>>> {
    Ok(Box::new(Id(10)))
}
```
The separate workaround can help us specify inner type, and return value can be cast to specific type successfully. Simple usage as follows.
```rust
    #[test]
    fn test_allocate() -> Result<(), Error> {
        let mut sys = SystemAllocator::new(); // Arc<Mutex<>> protection outside in vmm
        let id_allocator = IdAllocator::new(Id(1), Id(100)).ok_or(Error::Exist)?;
        let id_name = id_allocator.name();
        sys.add_uint_allocator(id_name.clone(), Box::new(id_allocator));

        // Here simply call directly
        let id = sys.allocate_uint(id_name.clone())?;
        assert_eq!(id.raw_value(), 10);
        Ok(())
    }
```

Although this makes SystemAllocator ugly to manually check which kind of resource it should call, it helps solve many problems above. 

What do you think?

andreeaflorescu:
@liujing2 I think we should first separate vm-allocator from vm-device and design them independently.

For the vm-allocator instead of exporting a `SystemAllocator` structure I would see it more like a trait, but I would first need to run some more experiments to see how would that play out. The `SystemAllocator` is then implemented by each VMM (Firecracker, Cloud Hypervisor, CrosVM, others) because only the VMM knows exactly what resources it will manage.

In rust-vmm/vm-allocator
```rust
pub trait SystemAllocator {
   pub fn add_allocator(...);
   pub fn allocate(...);
}
```

In Cloud Hypervisor:
```rust
struct VmmSystemAllocator {
   io_address_allocator: Arc<Mutex<IoAddressAllocator>>,
   irq_allocator: ....
```

This would make it easier because then the SystemAllocator doesn't need to be generic and work with HashMap or Vec of ResourceAllocator.

I am doing some prototypes and will keep you posted. If we separate the vm-allocator from the vm-device we should be able to keep working on the vm-device and maybe merge it sooner?

andreeaflorescu:
I have created a very rough prototype. If people like the idea I'll continue to work on it an improve it, but for now it doesn't even build.

The vm-allocator has the following traits:
- `ResourceAllocator` which uses a generic type that implements `Resource`
- `Resource` just a dummy empty trait used to make `ResourceAllocator` generic

And the following implementations:
- `Id` implements `Resource` and it's used by `IdResourceAllocator`
- `IdResourceAllocator` implements `ResourceAllocator` and basically is used for allocating u32 values which are wrapped in an `Id`.

My view of the vm-allocator would be to have a few more `ResourceAllocator` implementations like an allocator for guest memory (which would work with addresses) and other similar allocators (like MMIO?).

The VMM implementations would define the `SystemAllocator` as they see fit for their usecase. The code in `SystemAllocator` should be rather simple: some defines of allocators wrapped in Arc<Mutex>> (or not in the case of Firecracker) and wrappers over allocate() functions exported by vm-allocator.

https://github.com/andreeaflorescu/vm-allocator

jiangliu:
> I have created a very rough prototype. If people like the idea I'll continue to work on it an improve it, but for now it doesn't even build.
> 
> The vm-allocator has the following traits:
> 
> * `ResourceAllocator` which uses a generic type that implements `Resource`
> * `Resource` just a dummy empty trait used to make `ResourceAllocator` generic
> 
> And the following implementations:
> 
> * `Id` implements `Resource` and it's used by `IdResourceAllocator`
> * `IdResourceAllocator` implements `ResourceAllocator` and basically is used for allocating u32 values which are wrapped in an `Id`.
> 
> My view of the vm-allocator would be to have a few more `ResourceAllocator` implementations like an allocator for guest memory (which would work with addresses) and other similar allocators (like MMIO?).
> 
> The VMM implementations would define the `SystemAllocator` as they see fit for their usecase. The code in `SystemAllocator` should be rather simple: some defines of allocators wrapped in Arc> (or not in the case of Firecracker) and wrappers over allocate() functions exported by vm-allocator.
> 
> https://github.com/andreeaflorescu/vm-allocator

Thanks for quick PoC:)
It's more flexible by defining Traits and default implementations. But need more think about the way to declare allocation conditions(alginment, min, max etc).

liujing2:
> It's more flexible by defining Traits and default implementations. But need more think about the way to declare allocation conditions(alginment, min, max etc).

Yes, agree with @jiangliu . We need take care of not only allocation conditions, but also allocation request (a non-specific allocation request or a specific resource request).

andreeaflorescu:
@jiangliu @liujing2 let's gather a list of requirements, preferably in the issue that was introducing the vm-allocator crate. We can than make a more informed decision about the design. On a high level, I would still like to have the functionality of the vm-allocator exported as both Trait and implementations.

liujing2:
@andreeaflorescu 
One thing I'd like to explain is current PR doesn't mean to have vm-allocator crate inside vm-device crate, but they have strong dependency so one PR for easy review. Or else, I'd like to submit them as two crates' PRs instead :)

Considering your latest proposal, we need more think about the way vm-device::DeviceManager use the SystemAllocator trait or struct. 

- I am not sure what is the use of `SystemAllocator trait` and how its `allocate()` function implements. The only use I can imagine is for `DeviceManager`, and this is good. But I think a simple struct would help do either. 
But for `allocate()`, each resources can't use a simply function if we don't put them into one container (e.g. hashmap or vec). Or else, several functions and callback need to have for `DeviceManager`.

So I guess some of my previous experiments and codes basing on your trait design and also with some modifications would make some sense. I'd like to post it here https://github.com/liujing2/vm-device/commit/6758deede187d33e1c62a2650ac5ff85981fbdb8,  and wish that would provide some help to us.

- Another comment is, later when people have other system level resources demand, vm-allocator still needs some new patches implementing that and might also in vm-device.

liujing2:
> @jiangliu @liujing2 let's gather a list of requirements, preferably in the issue that was introducing the vm-allocator crate. We can than make a more informed decision about the design. On a high level, I would still like to have the functionality of the vm-allocator exported as both Trait and implementations.

Thanks for your proposal. It would be great to open a vm-allocator crate repo and discuss there. I'd like to post the PR again there for discussion.

liujing2:
Trying to list some big differences between very simple vmm (like firecracker) and general vmm request after some talk with @andreeaflorescu . If there's something misunderstanding, please correct me. :)

vm-allocator:
- All resources need to be managed (e.g. by SystemAllocator), recording which region is used and for what in general vmm. Simple vmm would not to have this.
- Each resource are thread safe. Who protecting this depends. e.g. `Arc<Mutex<IdAllocator>>` in SystemAllocator or VMM.
- Can provide two level APIs for simple vmm like firecraker and normal one: System level APIs and low level APIs on each resource type.

vm-device:
- Key to couple or decouple with vm-allocator: who is better responsible for each device resource allocation and free? VMM or DeviceManager?

andreeaflorescu:
@liujing2 can I suggest that you split the functionality in vm-allocator in multiple PRs to make it easier to review? We can have:
- one PR with trait definitions. The purpose of this PR would be to present the high level design of the crate. If we were to keep the design that I proposed, this PR would only have the following traits: Resource and ResourceAllocator.
- one PR with a minimal implementation of the trait. This one would have an implementation for IdAllocator and maybe AddressAllocator.
 
What do you say?

sameo:
@liujing2 I think we should close that PR now.

PR #6, opened 2019-08-05
DO_NOT_MERGE: a draft design for interrupt management

jiangliu:
This PR is a draft for interrupt management, which aims to decouple the vm-allocator crate from the vm-device crate.
@sboeuf @liujing2 @sameo @andreeaflorescu Please help to take a look at the proposal:)

### DIFF
--- a/src/interrupt/mod.rs
+++ b/src/interrupt/mod.rs
@@ -1,0 +1,188 @@
+//! Traits and Structs to manage interrupt sources for devices.
+//!
+//! In system programming, an interrupt is a signal to the processor emitted by hardware or
+//! software indicating an event that needs immediate attention. An interrupt alerts the processor
+//! to a high-priority condition requiring the interruption of the current code the processor is
+//! executing. The processor responds by suspending its current activities, saving its state, and
+//! executing a function called an interrupt handler (or an interrupt service routine, ISR) to deal
+//! with the event. This interruption is temporary, and, after the interrupt handler finishes,
+//! unless handling the interrupt has emitted a fatal error, the processor resumes normal
+//! activities.
+//!
+//! Hardware interrupts are used by devices to communicate that they require attention from the
+//! operating system, or a bare-metal program running on the CPU if there are no OSes. The act of
+//! initiating a hardware interrupt is referred to as an interrupt request (IRQ). Different devices
+//! are usually associated with different interrupts using a unique value associated with each
+//! interrupt. This makes it possible to know which hardware device caused which interrupts.
+//! These interrupt values are often called IRQ lines, or just interrupt lines.
+//!
+//! Nowadays, IRQ lines is not the only mechanism to delivery device interrupts to processors.
+//! MSI [(Message Signalled Interrupt)](https://en.wikipedia.org/wiki/Message_Signaled_Interrupts)
+//! is another commonly used alternative in-band method of signaling an interrupt, using special
+//! in-band messages to replace traditional out-of-band assertion of dedicated interrupt lines.
+//! While more complex to implement in a device, message signaled interrupts have some significant
+//! advantages over pin-based out-of-band interrupt signaling. Message signaled interrupts are
+//! supported in PCI bus since its version 2.2, and in later available PCI Express bus. Some non-PCI
+//! architectures also use message signaled interrupts.
+//!
+//! On the other hand, IRQ is a term commonly used by Operating Systems when dealing with hardware
+//! interrupts. The IRQ number managed by OSes is independent of the IRQ number managed by VMM.
+//! To make things easy, `Interrupt Source` is used instead of IRQ for pin-based interrupts and
+//! MSI interrupts.
+//!
+//! A device may support multiple types of interrupts, and each type of interrupt may support one
+//! or multiple interrupt sources. For example, a PCI device may support:
+//! * Legacy Irq: exactly one interrupt source.
+//! * PCI MSI Irq: 1,2,4,8,16,32 interrupt sources.
+//! * PCI MSIx Irq: 2^n(n=0-11) interrupt sources.
+//!
+//! A distinct Interrupt Source Identifier (ISID) will be assigned to each interrupt source.
+//! An ID allocator will be used to allocate and free Interrupt Source Identifiers for devices.
+//! To decouple the vm-device crate from the ID allocator, the vm-device crate doesn't take the
+//! responsibility to allocate/free Interrupt Source IDs but only makes use of assigned IDs.
+//!
+//! The overall flow to deal with interrupts is:
+//! * the VMM creates an interrupt manager
+//! * the VMM creates a device manager, passing on an reference to the interrupt manager
+//! * the device manager passes on an reference to the interrupt manager to all registered devices
+//! * guest kernel loads drivers for virtual devices
+//! * guest device driver determines the type and number of interrupts needed, and update the
+//!   device configuration
+//! * the virtual device backend requests the interrupt manager to create an interrupt group
+//!   according to guest configuration information
+
+use std::sync::Arc;
+use vmm_sys_util::eventfd::EventFd;
+
+/// Reuse std::io::Result to simplify inter operability.
+pub type Result<T> = std::io::Result<T>;
+
+/// Data type to store an interrupt index.
+pub type InterruptIndex = u32;
+
+#[cfg(feature = "generic_msi")]
+/// Maximum number of generic Message Signalled Interrupts supported per device.
+pub const MAX_GENERIC_MSI_IRQS: InterruptIndex = 32;
+
+/// Type of interrupt source.
+pub enum InterruptSourceType {
+    #[cfg(feature = "legacy_irq")]
+    /// Legacy Pin-based Interrupt.
+    /// On x86 platforms, legacy interrupts are routed through 82599 PICs and/or IOAPICs.
+    Legacy,
+    #[cfg(feature = "generic_msi")]
+    /// Generic Message Signalled Interrupt.
+    /// Some non-PCI devices (like HPET on x86) make use of generic MSI in platform specific ways.
+    GenericMsi,
+    #[cfg(feature = "pci_msi")]
+    /// PCI Message Signalled Interrupt.
+    PciMsi,
+    #[cfg(feature = "pci_msix")]
+    /// PCI Message Signalled Interrupt - Second Generation.
+    PciMsix,
+}
+
+/// Configuration data for an interrupt source.
+pub enum InterruptSourceConfig {
+    /// Configuration data for Legacy interrupts.
+    Legacy(LegacyIrqSourceConfig),
+    /// Configuration data for GenericMsi, PciMsi, PciMsix interrupts.
+    Msi(MsiIrqSourceConfig),
+}
+
+/// Configuration data for legacy interrupts.
+///
+/// On x86 platforms, legacy interrupts means those interrupts routed through PICs or IOAPICs.
+pub struct LegacyIrqSourceConfig {
+    /// Bitmask for valid bits in the corresponding interrupt status register.
+    pub valid_flag_mask: u32,
+}
+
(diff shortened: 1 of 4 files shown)


andreeaflorescu:
P.S. you can create a draft PR. The draft PR cannot be merged. Check it out here: https://github.blog/2019-02-14-introducing-draft-pull-requests/

src/interrupt/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+//!
+//! The overall flow to deal with interrupts is:
+//! * the VMM creates an interrupt manager
+//! * the VMM creates a device manager, passing on an reference to the interrupt manager
+//! * the device manager passes on an reference to the interrupt manager to all registered devices
+//! * guest kernel loads drivers for virtual devices
+//! * guest device driver determines the type and number of interrupts needed, and update the
+//!   device configuration
+//! * the virtual device backend requests the interrupt manager to create an interrupt group
+//!   according to guest configuration information
```
liujing2: @jiangliu Thanks for the proposal.

The idea to make vm-device crate manage the interrupt resource and related methods is great.
One place not clear to me is, what does it mean by "device manager passes on an reference to the interrupt manager"? Do you mean that the device manager records the interrupt group that a device uses, e.g. `<LegacyIrq, Arc<Mutex<<dyn Device>>>`?

jiangliu: That means passing an argument of Arc<Box<dyn InterruptManager>> to the device manager, so the device manager could clone/call methods of InterruptManager trait object.

liujing2: Thanks. 
Do we also have request to let device manager get the information of which interrupt/number/group a device uses?
Some future request like dump a device information or something similar...

jiangliu: Aha, one point is missing in the proposal, how to allocate interrupt source identifier from the vm_allocator. 
I think pre-allocator enough interrupt source identifier for each device should work for most cases. For example, PCI MSIx supports 2048 interrupts, but 32 interrupts should work for most cases:)
Or we could determine number of interrupts needed when creating virtual device backend.

liujing2: For pin irq, virtual backend could specify a number or just let vm-allocator allocate one. For MSI/MSI-X, why we need preallocate?

jiangliu: >For pin irq, virtual backend could specify a number or just let vm-allocator allocate one. 
Per my understanding, any interrupt source id allocated from the vm-allocator should be OK.

>For MSI/MSI-X, why we need preallocate?
Otherwise when guest OS enabling MSI, the virtual device backend needs to allocate interrupt source IDs from the allocator, which forces dependency on the vm-allocator.

liujing2: > > For MSI/MSI-X, why we need preallocate?
> > Otherwise when guest OS enabling MSI, the virtual device backend needs to allocate interrupt source IDs from the allocator, which forces dependency on the vm-allocator.

Not sure if there're some misunderstandings. :)
The guest is the only role to allocate MSI vector for a device. And as we talked offline, there's no need to set back to allocator when virtual device backend traps a MSI data (vector inside). So there'll be no dependency.

jiangliu: By default, the kvm kernel module supports 1024 interrupt sources at most. For MSI, we still need to allocate a distinct interrupt source ID for for each MSI interrupt, so we could call KVM_SET_IRQ_ROUTING for setup interrupt route.

sboeuf: Well we have two options to make the dependency between KVM and MSI. Either we store MSI vector so that they can be called directly with `KVM_SIGNAL_MSI`, or we decide some more generic design relying on `KVM_SET_GSI_ROUTING` and using `EventFd` to notify about an interrupt and let KVM do the mapping between the file descriptor and the associated MSI.
For instance, in the context of VFIO, `KVM_SET_GSI_ROUTING` is very convenient since it avoids the interrupt generation coming from the device to go back to userspace. But on the other hand, `KVM_SET_GSI_ROUTING` has a drawback, the entire table has to be updated everytime an MSI vector is updated.

jiangliu: I think KVM_SET_GSI_ROUTING is the right way. KVM_SIGNAL_MSI is convenient for the VMM to inject interrupts into the guests, but it doesn't support posted interrupt and will cause penalty when dealing device pass through.
For KVM_SET_GSI_ROUTING, it depends on a global view of all interrupts, so we let the InterruptManager to deal with KVM_SET_GSI_ROUTING:)

src/interrupt/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Type of interrupt source.
+pub enum InterruptSourceType {
+    #[cfg(feature = "legacy_irq")]
+    /// Legacy Pin-based Interrupt.
+    /// On x86 platforms, legacy interrupts are routed through 82599 PICs and/or IOAPICs.
+    Legacy,
+    #[cfg(feature = "generic_msi")]
+    /// Generic Message Signalled Interrupt.
+    /// Some non-PCI devices (like HPET on x86) make use of generic MSI in platform specific ways.
```
sboeuf: Is `generic MSI` a real thing? I mean I know about very specific HPET use case, but is it standardized in any way? As a started, we should stick to the 3 other options, and leave the very specific HPET use case aside.

jiangliu: We are extending the Virtio MMIO transport layer to support MSI, so used "generic MSI" for it.
Or we may called it "MmioMsi", something similar to HPET, DMAR or platform MSI on arm.

src/interrupt/mod.rs
sboeuf: >For KVM_SET_GSI_ROUTING, it depends on a global view of all interrupts, so we let the InterruptManager to deal with KVM_SET_GSI_ROUTING:)

Easy to say :) and hopefully not too hard to put in practice. I'm only concerned about the fact that the `InterruptManager` will have to be aware about any MSI update coming from the guest. And those updates can come through BAR or PCI config writing, which means we need mechanisms to callback into the `InterruptManager` to let it know that something happened.
The part where the `InterruptManager` decides to update `KVM_SET_GSI_ROUTING` list based on its internal entries seems easy indeed :)

src/interrupt/mod.rs
sboeuf: Yes `MmioMsi` sounds a bit better to me, as it contrasts with PCI. But the thing is that you're going to call something `MmioMsi` when it's not part of a spec yet. I'm worried that we're mixing very well defined types with other ones that are not yet approved by any upstream, WDYT?

jiangliu: Sadly, that's the truth:(
How about first have some concrete implementation and benchmark data, then we could propose to update viritio spec.

sboeuf: Sounds good!

src/interrupt/mod.rs
liujing2: @jiangliu @sboeuf 

> > For KVM_SET_GSI_ROUTING, it depends on a global view of all interrupts, so we let the InterruptManager to deal with KVM_SET_GSI_ROUTING:)
> 
> Easy to say :) and hopefully not too hard to put in practice. I'm only concerned about the fact that the `InterruptManager` will have to be aware about any MSI update coming from the guest. And those updates can come through BAR or PCI config writing, which means we need mechanisms to callback into the `InterruptManager` to let it know that something happened.
> The part where the `InterruptManager` decides to update `KVM_SET_GSI_ROUTING` list based on its internal entries seems easy indeed :)

So we decide to *only* make vm-device::InterruptManager to know the global view of all interrupts (especially msi udpate), instead of have vm-allocator to know that? 

So for interrupt, can I say that the actual work for vm-allocator now is to just pre-allocate enough for each device during initialization?

sboeuf: That sounds right to me. Since the vm-allocator will never have to handle conflicts and/or allocations of MSI/MSI-X, the vm-device can be the one being notified about the values of those interrupt vectors only.

jiangliu: > > For instance, in the context of VFIO, KVM_SET_GSI_ROUTING is very convenient since it avoids the interrupt generation coming from the device to go back to userspace. But on the other hand, KVM_SET_GSI_ROUTING has a drawback, the entire table has to be updated everytime an MSI vector is updated.
> 
> Agreed. That's a drawback, but I think `KVM_SET_GSI_ROUTING` is a more generic scheme that allows us to support both VMM injection and posted interrupts.

We are prototyping by using KVM_SET_GSI_ROUTING because it's very friendly to POSTED_INTR.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+//! * PCI MSIx Irq: 2^n(n=0-11) interrupt sources.
+//!
+//! A distinct Interrupt Source Identifier (ISID) will be assigned to each interrupt source.
+//! An ID allocator will be used to allocate and free Interrupt Source Identifiers for devices.
+//! To decouple the vm-device crate from the ID allocator, the vm-device crate doesn't take the
+//! responsibility to allocate/free Interrupt Source IDs but only makes use of assigned IDs.
+//!
+//! The overall flow to deal with interrupts is:
+//! * the VMM creates an interrupt manager
+//! * the VMM creates a device manager, passing on an reference to the interrupt manager
```
sameo: For the sake of API simplicity, should we have the device manager creation API implicitly create the interrupt manager?
Do you need access to the interrupt manager from the VMM itself?

src/interrupt/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+//! * the device manager passes on an reference to the interrupt manager to all registered devices
+//! * guest kernel loads drivers for virtual devices
+//! * guest device driver determines the type and number of interrupts needed, and update the
+//!   device configuration
+//! * the virtual device backend requests the interrupt manager to create an interrupt group
```
sameo: I'm trying to understand how you think this would interact with the [proposed device manager design](https://github.com/rust-vmm/vm-allocator/pull/1#issuecomment-518003480) and update it accordingly:

1. When registering a device, would interrupts be represented as resources at all?

2. With legacy IRQs, the VMM selects the interrupt numbers the device is going to use. Is GSI allocation part of the vm-allocator crate?

3. With MSI, the addresses that the guest chooses are part of the MMIO address space. Do we let the vm-allocator crate know about it, or do we assume there won't be any conflict with other devices where the VMM needs to reserve allocate chunks of the MMIO address space?

Looking forward to getting your input, but my personal answers:

1. Eventually, we will need to have the device manager know about interrupts for any device. In particular for legacy interrupts. Maybe we should only consider legacy interrupts as a device manager resource.

2. I think GSI should be managed by the vm-allocator.

3. That may be part of the next step for updating a device's resources, but here as well the device manager will need to be aware of which MSI addresses have been selected by the guest, and not allocate anything there.

sameo: > That may be part of the next step for updating a device's resources, but here as well the device manager will need to be aware of which MSI addresses have been selected by the guest, and not allocate anything there.

@chao-p rightfully pointed out that MSI addresses are only used by KVM for resolving the target interrupt. They will not be used for actual IO by the guest. So I guess we can scratch point #3 off.

jiangliu: Are you talking about the high_address/low_address in PCI MSI configuration registers? That address must fall into the local APIC address range(0xFEE000000), which is reserved by the VMMs.

src/interrupt/mod.rs
jiangliu: It turns out current design works perfectly:)

sameo: > It turns out current design works perfectly:)

You mean the KVM_SET_GSI_ROUTING based design? That's great, yes.

sameo:
@jiangliu Can we close this one as it's replaced by #8 ?

jiangliu:
Sure, I will address comments and then create a formal PR.

PR #12, opened 2019-10-31
Add IO manager support

liujing2:
Let's firstly add IO manager which focuses on device Port IO and MMIO management, and static IO mapping in runtime as the first step. 
@sameo @andreeaflorescu  @jiangliu  PTAL, thanks.

### DIFF
--- a/src/device_manager.rs
+++ b/src/device_manager.rs
@@ -12,8 +12,9 @@
 //! devices IO ranges, and finally set resources to virtual device.
 
 use crate::resources::Resource;
-use crate::DeviceIo;
+use crate::{DeviceIo, IoAddress, IoSize};
 
+use std::cmp::{Ord, Ordering, PartialEq, PartialOrd};
 use std::collections::btree_map::BTreeMap;
 use std::result;
 use std::sync::Arc;
@@ -23,18 +24,62 @@ use std::sync::Arc;
 pub enum Error {
     /// The inserting device overlaps with a current device.
     DeviceOverlap,
+    /// The device doesn't exist.
+    NoDevice,
 }
 
 /// Simplify the `Result` type.
 pub type Result<T> = result::Result<T, Error>;
 
+// Structure describing an IO range.
+#[derive(Debug, Copy, Clone)]
+struct IoRange {
+    base: IoAddress,
+    size: IoSize,
+}
+
+impl IoRange {
+    fn new_pio_range(base: u16, size: u16) -> Self {
+        IoRange {
+            base: IoAddress::Pio(base),
+            size: IoSize::Pio(size),
+        }
+    }
+    fn new_mmio_range(base: u64, size: u64) -> Self {
+        IoRange {
+            base: IoAddress::Mmio(base),
+            size: IoSize::Mmio(size),
+        }
+    }
+}
+
+impl Eq for IoRange {}
+
+impl PartialEq for IoRange {
+    fn eq(&self, other: &IoRange) -> bool {
+        self.base == other.base
+    }
+}
+
+impl Ord for IoRange {
+    fn cmp(&self, other: &IoRange) -> Ordering {
+        self.base.cmp(&other.base)
+    }
+}
+
+impl PartialOrd for IoRange {
+    fn partial_cmp(&self, other: &IoRange) -> Option<Ordering> {
+        self.base.partial_cmp(&other.base)
+    }
+}
+
 /// System IO manager serving for all devices management and VM exit handling.
 #[derive(Default)]
 pub struct IoManager {
     /// Range mapping for VM exit pio operations.
-    pio_bus: BTreeMap<(u16, u16), Arc<dyn DeviceIo>>,
+    pio_bus: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
     /// Range mapping for VM exit mmio operations.
-    mmio_bus: BTreeMap<(u64, u64), Arc<dyn DeviceIo>>,
+    mmio_bus: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
 }
 
 impl IoManager {
@@ -60,7 +105,11 @@ impl IoManager {
         for (idx, res) in resources.iter().enumerate() {
             match *res {
                 Resource::PioAddressRange { base, size } => {
-                    if self.pio_bus.insert((base, size), device.clone()).is_some() {
+                    if self
+                        .pio_bus
+                        .insert(IoRange::new_pio_range(base, size), device.clone())
+                        .is_some()
+                    {
                         // Unregister registered resources.
                         self.unregister_device_io(&resources[0..idx])
                             .expect("failed to unregister devices");
@@ -69,7 +118,11 @@ impl IoManager {
                     }
                 }
                 Resource::MmioAddressRange { base, size } => {
-                    if self.mmio_bus.insert((base, size), device.clone()).is_some() {
+                    if self
+                        .mmio_bus
+                        .insert(IoRange::new_mmio_range(base, size), device.clone())
+                        .is_some()
+                    {
                         // Unregister registered resources.
                         self.unregister_device_io(&resources[0..idx])
                             .expect("failed to unregister devices");
@@ -95,14 +148,89 @@ impl IoManager {
         for res in resources.iter() {
             match *res {
                 Resource::PioAddressRange { base, size } => {
-                    self.pio_bus.remove(&(base, size));
+                    self.pio_bus.remove(&IoRange::new_pio_range(base, size));
                 }
                 Resource::MmioAddressRange { base, size } => {
-                    self.mmio_bus.remove(&(base, size));
+                    self.mmio_bus.remove(&IoRange::new_mmio_range(base, size));
                 }
                 _ => continue,
             }
         }
         Ok(())
+    }
+
+    fn get_entry(&self, addr: IoAddress) -> Option<(&IoRange, &Arc<dyn DeviceIo>)> {
+        match addr {
+            IoAddress::Pio(a) => self
+                .pio_bus
+                .range(..=&IoRange::new_pio_range(a, 0))
+                .nth_back(0),
+            IoAddress::Mmio(a) => self
+                .mmio_bus
+                .range(..=&IoRange::new_mmio_range(a, 0))
+                .nth_back(0),
+        }
+    }
+
+    // Return the Device mapped `addr` and the base address.
+    fn get_device(&self, addr: IoAddress) -> Option<(&Arc<dyn DeviceIo>, IoAddress)> {
+        if let Some((range, dev)) = self.get_entry(addr) {
+            if (addr.raw_value() - range.base.raw_value()) < range.size.raw_value() {
+                return Some((dev, range.base));
+            }
+        }
+        None
+    }
+
+    /// A helper function handling PIO read command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn pio_read(&self, addr: u16, data: &mut [u8]) -> Result<()> {
+        if let Some((device, base)) = self.get_device(IoAddress::Pio(addr)) {
+            device.read(base, IoAddress::Pio(addr - (base.raw_value() as u16)), data);
+            Ok(())
+        } else {
+            Err(Error::NoDevice)
+        }
+    }
+
+    /// A helper function handling PIO write command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn pio_write(&self, addr: u16, data: &[u8]) -> Result<()> {
+        if let Some((device, base)) = self.get_device(IoAddress::Pio(addr)) {
+            device.write(base, IoAddress::Pio(addr - (base.raw_value() as u16)), data);
+            Ok(())
+        } else {
+            Err(Error::NoDevice)
+        }
+    }
+
+    /// A helper function handling MMIO read command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn mmio_read(&self, addr: u64, data: &mut [u8]) -> Result<()> {
+        if let Some((device, base)) = self.get_device(IoAddress::Mmio(addr)) {
+            device.read(base, IoAddress::Mmio(addr - base.raw_value()), data);
+            Ok(())
+        } else {
+            Err(Error::NoDevice)
+        }
+    }
+
+    /// A helper function handling MMIO write command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn mmio_write(&self, addr: u64, data: &[u8]) -> Result<()> {
+        if let Some((device, base)) = self.get_device(IoAddress::Mmio(addr)) {
+            device.write(base, IoAddress::Mmio(addr - base.raw_value()), data);
+            Ok(())
+        } else {
+            Err(Error::NoDevice)
+        }
     }
 }
--- a/src/device_manager.rs
+++ b/src/device_manager.rs
@@ -234,3 +234,125 @@ impl IoManager {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::Mutex;
+
+    const PIO_ADDRESS_SIZE: u16 = 4;
+    const PIO_ADDRESS_BASE: u16 = 0x40;
+    const MMIO_ADDRESS_SIZE: u64 = 0x8765_4321;
+    const MMIO_ADDRESS_BASE: u64 = 0x1234_5678;
+    const LEGACY_IRQ: u32 = 4;
+    const CONFIG_DATA: u32 = 0x1234;
+
+    struct DummyDevice {
+        config: Mutex<u32>,
+    }
+
+    impl DummyDevice {
+        fn new(config: u32) -> Self {
+            DummyDevice {
+                config: Mutex::new(config),
+            }
+        }
+    }
+
+    impl DeviceIo for DummyDevice {
+        fn read(&self, _base: IoAddress, _offset: IoAddress, data: &mut [u8]) {
+            if data.len() > 4 {
+                return;
+            }
+            for (idx, iter) in data.iter_mut().enumerate() {
+                let config = self.config.lock().expect("failed to acquire lock");
+                *iter = (*config >> (idx * 8) & 0xff) as u8;
+            }
+        }
+
+        fn write(&self, _base: IoAddress, _offset: IoAddress, data: &[u8]) {
+            let mut config = self.config.lock().expect("failed to acquire lock");
+            *config = u32::from(data[0]) & 0xff;
+        }
+    }
+
+    #[test]
+    fn test_register_unregister_device_io() {
+        let mut io_mgr = IoManager::new();
+        let dummy = DummyDevice::new(0);
+        let dum = Arc::new(dummy);
+
+        let mut resource: Vec<Resource> = Vec::new();
+        let mmio = Resource::MmioAddressRange {
+            base: MMIO_ADDRESS_BASE,
+            size: MMIO_ADDRESS_SIZE,
+        };
+        let irq = Resource::LegacyIrq(LEGACY_IRQ);
+
+        resource.push(mmio);
+        resource.push(irq);
+
+        assert!(io_mgr.register_device_io(dum.clone(), &resource).is_ok());
+        assert!(io_mgr.unregister_device_io(&resource).is_ok())
+    }
+
+    #[test]
+    fn test_mmio_read_write() {
+        let mut io_mgr: IoManager = Default::default();
+        let dum = Arc::new(DummyDevice::new(CONFIG_DATA));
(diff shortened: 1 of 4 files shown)


sameo:
I'm marking that one as WIP for now.

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+        Ok(())
+    }
+
+    fn first_before(
+        &self,
+        addr: IoAddress,
+        map: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
+    ) -> Option<(IoRange, Arc<dyn DeviceIo>)> {
+        for (range, dev) in map.iter().rev() {
```
jiangliu: We should use BtreeMap::range()(https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) here, it's available since rust 1.17 and should be more efficient than iter().

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+        None
+    }
+
+    // Return the Device mapped the address.
+    fn get_device(
+        &self,
+        addr: IoAddress,
+        map: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
```
jiangliu: Why taking ownership instead of taking a reference here?

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+        None
+    }
+
+    /// A helper function handling PIO read command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn pio_read(&self, addr: u16, data: &mut [u8]) -> Result<()> {
+        if let Some(device) = self.get_device(IoAddress::Pio(addr), self.pio_bus.clone()) {
```
jiangliu: The IoManager::pio_bus() is an instance of BtreeMap, the clone() implies a deep copy. It's really really too heavy here.
Please refer to Clone for BtreeMap:
https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#138-214

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    /// A helper function handling MMIO read command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn mmio_read(&self, addr: u64, data: &mut [u8]) -> Result<()> {
+        if let Some(device) =
+            self.get_device(IoAddress::Mmio(GuestAddress(addr)), self.mmio_bus.clone())
```
jiangliu: Ditto.

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    /// A helper function handling MMIO write command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn mmio_write(&self, addr: u64, data: &[u8]) -> Result<()> {
+        if let Some(device) =
+            self.get_device(IoAddress::Mmio(GuestAddress(addr)), self.mmio_bus.clone())
+        {
```
jiangliu: Ditto

src/device_manager.rs
```diff
@@ -0,0 +1,365 @@
...
+        } else {
+            Err(Error::DeviceNotExist)
+        }
+    }
+
+    /// A helper function handling PIO write command during VM exit.
+    /// The virtual device itself provides mutable ability and thead-safe protection.
+    ///
+    /// Return error if failed to get the device.
+    pub fn pio_write(&self, addr: u16, data: &[u8]) -> Result<()> {
```
jiangliu: Ditto

src/device_manager.rs
liujing2: > We should use BtreeMap::range()(https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) here, it's available since rust 1.17 and should be more efficient than iter().

Updated with new method.

src/device_manager.rs
liujing2: Solved and removed.

src/lib.rs
```diff
@@ -19,15 +41,47 @@ pub enum IoAddress {
...
+        self.raw_value().partial_cmp(&other.raw_value())
+    }
+}
+
 /// Device IO trait.
 /// A device supporting memory based I/O should implement this trait, then
 /// register itself against the different IO type ranges it handles.
 /// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
 /// registered devices read or write method from this trait.
+/// The mutable ability is provided by virtual device inside. e.g. Mutex, RwLock.
```
andreeaflorescu: How do we implement simple devices such as the Serial Console if we don't have mutability of self?

You can check this PR to see what I mean: https://github.com/rust-vmm/vm-device/pull/15

liujing2: > How do we implement simple devices such as the Serial Console if we don't have mutability of self?
> 
> You can check this PR to see what I mean: #15

We need protect the thread-safe by e.g. Mutex, RwLock behind the devices. So we could get mutability inside devices, after getting the corresponding `dyn DeviceIo` by address.

src/device_manager.rs
liujing2: It seems our CI doesn't accept `std::iter::DoubleEndedIterator` trait which is implemented for `Range` used by `BtreeMap::range()`? @sameo @jiangliu

src/lib.rs
andreeaflorescu: Why is this better than having `&mut`? I think devices by design mutate their internal state when a read/write occurs so you might as well just be explicit about this by using &mut self.

liujing2: This is designed for multiple threads accessing different IO windows on one device.

liujing2: > Why is this better than having `&mut`? I think devices by design mutate their internal state when a read/write occurs so you might as well just be explicit about this by using &mut self.

Multiple threads accessing different IO windows on one device would benefit from the new one.

jiangliu: The main idea is to support concurrent access from multiple vCPUs, which is important for large scale instances.

src/device_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+#[derive(Default)]
+pub struct IoManager {
+    /// Range mapping for VM exit pio operations.
+    pio_bus: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
+    /// Range mapping for VM exit mmio operations.
+    mmio_bus: BTreeMap<IoRange, Arc<dyn DeviceIo>>,
+}
+
+impl IoManager {
+    pub fn new() -> Self {
```
jiangliu: Missing docs?

src/device_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+        Ok(())
+    }
+
+    // BTreeMap::range() yields elements from min(inclusive) to max(exclusive).
+    // Let's use a fake range for getting inclusive result.
+    fn get_entry(&self, addr: IoAddress) -> Option<(&IoRange, &Arc<dyn DeviceIo>)> {
+        match addr {
+            IoAddress::Pio(a) => self
+                .pio_bus
+                .range(..&IoRange::new_pio_range(a + 1, 0))
```
jiangliu: what happens if addr is Pio(0xffff)? Seems it may cause overflow.

src/device_manager.rs
```diff
@@ -0,0 +1,360 @@
...
+                .range(..&IoRange::new_mmio_range(a + 1, 0))
+                .nth_back(0),
+        }
+    }
+
+    // Return the Device mapped the address.
+    fn get_device(&self, addr: IoAddress) -> Option<Arc<dyn DeviceIo>> {
+        if let Some((range, dev)) = self.get_entry(addr) {
+            if (addr.raw_value() - range.base.raw_value()) < range.size.raw_value() {
+                return Some(dev.clone());
```
jiangliu: Arc::clone() is still a relative heavy operations and DeviceIo::{read, write} takes an &self. So could we return &Arc<dyn DeviceIo> instead of Arc<dyn DeviceIo> here?

src/lib.rs
```diff
@@ -5,10 +5,32 @@
...
+    /// Memory mapped I/O size.
+    Mmio(GuestUsize),
+}
+
+impl IoSize {
+    /// Get the raw value as u64 to make operation simple.
+    pub fn raw_value(&self) -> u64 {
+        match *self {
+            IoSize::Pio(p) => u64::from(p),
+            IoSize::Mmio(m) => m,
```
jiangliu: This assumes that GuestUsize is u64:)

liujing2: > This assumes that GuestUsize is u64:)

It seems that vm-memory definition shows GuestUsize is u64.
``` rust
impl_address_ops!(GuestAddress, u64);

macro_rules! impl_address_ops {
    ($T:ident, $V:ty) => {
        impl AddressValue for $T {
            type V = $V;
        }

```
```
pub type GuestUsize = <GuestAddress as AddressValue>::V;
```

jiangliu: That's an vm-memory internal detail, it would be better to rely on that.

jiangliu: Actually I'm hesitating whether we should remove dependency on the vm-memory crate by using u64 instead of GuestAddress/GuestUsize.
What do you think?

liujing2: > Actually I'm hesitating whether we should remove dependency on the vm-memory crate by using u64 instead of GuestAddress/GuestUsize.
> What do you think?

When we considering using vm-memory::GuestAddress/GuestUsize or not, it means two things: whether IoManager knows the mmio address type, and whether device IO interface (`DeviceIo::read/write`) and device inside knows the mmio address type (for resource requirement and assignment). It would be better to keep those the same.

jiangliu: > > Actually I'm hesitating whether we should remove dependency on the vm-memory crate by using u64 instead of GuestAddress/GuestUsize.
> > What do you think?
> 
> When we considering using vm-memory::GuestAddress/GuestUsize or not, it means two things: whether IoManager knows the mmio address type, and whether device IO interface (`DeviceIo::read/write`) and device inside knows the mmio address type (for resource requirement and assignment). It would be better to keep those consistent.

So we need to decide the argument to DeviceIo::{read,write} is an address or an offset.
I suggest to use offset, because that will simplify the driver implementations.

andreeaflorescu: I also prefer the argument to deviceio to be an offset. In which case I would also propose to change the signature of read/write to use `offset: u64` instead of `addr: IoAddress`.

jiangliu: > I also prefer the argument to deviceio to be an offset. In which case I would also propose to change the signature of read/write to use `offset: u64` instead of `addr: IoAddress`.

We still need a way to encode the access type: PIO vs MMIO. So we either add an extra `access_type` parameter or use the IoAddress enum. I suggest to use IoAddress enum to encode the access type. The reason is that PCI device need to support both PIO and MMIO.

liujing2: > We still need a way to encode the access type: PIO vs MMIO. So we either add an extra `access_type` parameter or use the IoAddress enum. 

Agree. In my very previous PR https://github.com/rust-vmm/vm-device/pull/3/commits/538836161c3dddb37f268ca41e1fa565ebfa7492,  there proposed such a type differing pio and mmio. 
> I suggest to use IoAddress enum to encode the access type. The reason is that PCI device need to support both PIO and MMIO.

Encoding it into address type is a good way to simply interface. However, once we choose current `IoAddress` enum for `DeviceIo::read/write`, it means `vm_device::IoManager` cares `GuestAddress/GuestUsize` instead of direct type u64, and transports such semantics to virtual device. So virtual device has semantics of `GuestAddress/GuestUsize` instead of direct u64 for IO windows, which means `ResourceConstraint` and `Resource` also need that semantics too.  That is what I mean by keeping consistency.

One simple way is, only use u64 instead of `GuestAddress`, and encode in enum as `offset` for IoManager::read/write.
```rust
/// IO Addresses.
#[derive(Debug, Copy, Clone)]
pub enum IoAddress {
    /// Port I/O address.
    Pio(u16),

    /// Memory mapped I/O address.
    Mmio(u64),
}
```

jiangliu: I like this design, make vm-device independent of vm-memory.

liujing2:
Could @sameo @andreeaflorescu @jiangliu take a look if there are any more comments? Thanks.

sboeuf:
@liujing2 you need to fix the CI by replacing `nth_back()` function with something that is not considered `unstable`.

src/device_manager.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/// Simplify the `Result` type.
+pub type Result<T> = result::Result<T, Error>;
+
+/// System IO manager serving for all devices management and VM exit handling.
+#[derive(Default)]
+pub struct IoManager {
+    /// Range mapping for VM exit pio operations.
+    pio_bus: BTreeMap<(u16, u16), Arc<dyn DeviceIo>>,
```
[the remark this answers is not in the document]
liujing2: Yes, let's keep it as Arc now since it can achieve real multi-thread handling as discussed  https://github.com/rust-vmm/vm-device/pull/12#discussion_r350146303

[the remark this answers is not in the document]
sboeuf: Yes of course, `Arc` is fine, and I agree this allows multi-threading, but at some point, we might need two different thread to modify the instance. That's where you need `Mutex` to be added to the picture.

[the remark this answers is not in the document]
jiangliu: The idea is to move lock into the `dyn DeviceIo` object, and let the device driver make decision how to protect itself from concurrent access.

liujing2:
> @liujing2 you need to fix the CI by replacing `nth_back()` function with something that is not considered `unstable`.

Thanks very much for the review.
I noticed the github rust issue directed by CI result, saying that `nth_back()` is stabilized in 1.37.0, see
https://github.com/rust-lang/rust/issues/56995#issuecomment-552668905
That's why I locally have no error. Do we need to update here? Or else, we need use complex handling in code.

sboeuf:
@liujing2 ah ok then I agree we should bump the CI to the latest Rust version.

sameo:
@liujing2 Could you please force push to check if CI passes with the new container?

andreeaflorescu:
@sameo We need to update the rust-vmm-ci git submodule so that the tests are run with the latest container. The changes are not picked up on new pushes on purpose so that new container versions don't break the build of a repository with no code changes.

sameo:
@liujing2 The tests are passing now, only the coverage check is failing.

liujing2:
> @liujing2 The tests are passing now, only the coverage check is failing.

@sameo Thanks for the help. It is successful now.

src/device_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+        }
+    }
+
+    // Return the Device mapped the address.
+    fn get_device(&self, addr: IoAddress) -> Option<&Arc<dyn DeviceIo>> {
+        if let Some((range, dev)) = self.get_entry(addr) {
+            if (addr.raw_value() - range.base.raw_value()) < range.size.raw_value() {
+                return Some(dev);
+            }
+            return None;
```
jiangliu: This line could be removed.

src/lib.rs
```diff
@@ -19,15 +41,47 @@ pub enum IoAddress {
...
+        self.raw_value().partial_cmp(&other.raw_value())
+    }
+}
+
 /// Device IO trait.
 /// A device supporting memory based I/O should implement this trait, then
 /// register itself against the different IO type ranges it handles.
 /// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
 /// registered devices read or write method from this trait.
+/// The mutable ability is provided by virtual device inside. e.g. Mutex, RwLock.
```
jiangliu: How about 
"The DeviceIo trait adopts the interior mutability pattern"

src/device_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+        for (idx, res) in resources.iter().enumerate() {
+            match *res {
+                Resource::PioAddressRange { base, size } => {
+                    if self
+                        .pio_bus
+                        .insert(IoRange::new_pio_range(base, size), device.clone())
+                        .is_some()
+                    {
+                        // Unregister registered resources.
+                        self.unregister_device_io(&resources[0..idx])?;
```
jiangliu: Should we use expect() or unwrap() here?
The unregister_device_io() shouldn't fail here, otherwise we are already in troubles.

src/lib.rs
sameo: > I like this design, make vm-device independent of vm-memory.

I like it as well. That combined with replacing the current `addr` argument with `base + offset` (see https://github.com/rust-vmm/vm-device/pull/13/files#r364223789) could make it into the same PR, to be merged before that one.

liujing2:
> Overall, the PR looks good.
> I would appreciate if we could separate the IO ops internal mutability changes into a separate commit, and also remove the vm-memory dependency (and split the `addr` argument into `base, offset`) from vm-device prior to merging that PR.

The difference between passing `base` `offset` and passing `addr` will be only for non-pci device instance.  These device instance can use `offset` directly to handle r/w emulation. But for those pci device instances, especially with several windows, they still need calculate `addr(base+offset)` and compare with each BAR Base Address to find the right one, right?

jiangliu:
> > Overall, the PR looks good.
> > I would appreciate if we could separate the IO ops internal mutability changes into a separate commit, and also remove the vm-memory dependency (and split the `addr` argument into `base, offset`) from vm-device prior to merging that PR.
> 
> Sure, we can separate the PR into three.
> 
> 1. remove vm-memory dependency; find a way for the DeviceIO read/write parameters.
> 2. Io ops internal mutability changing.
> 3. IoManager related.
> 
> For the 1st, the difference between passing `base` `offset` and passing `addr` will be only for non-pci device instance, I think. These device instance can use `offset` directly to handle r/w emulation. But for those pci device instances, especially with several windows, they still need calculate `addr(base+offset)` and compare with each BAR Base Address to find the right one, right?

If 'base + offset' is used, the driver may need to lookup some internal data structures by using base as the key. And `token` actually caches information about the internal data structures thus avoid the looking up operations.

sameo:
> If 'base + offset' is used, the driver may need to lookup some internal data structures by using base as the key. And `token` actually caches information about the internal data structures thus avoid the looking up operations.

Yes, that's right that the device may do some lookup, but this could easily be a O(1) lookup for those devices that actually need to look it up (PCI devices, yes). This seems to me like a micro optimization at that point, and the cost for it is a bound between the IO trait and the IO manager. In other words, callers of the IO trait methods must be stateful by keeping track of a cookie/token. Using this kind of API builds an implicit dependency between the trait and a registration method. To me, that's not the cleanest design, and the benefit we'd get for it is arguable (an O(1) lookup).

jiangliu:
> > If 'base + offset' is used, the driver may need to lookup some internal data structures by using base as the key. And `token` actually caches information about the internal data structures thus avoid the looking up operations.
> 
> Yes, that's right that the device may do some lookup, but this could easily be a O(1) lookup for those devices that actually need to look it up (PCI devices, yes). This seems to me like a micro optimization at that point, and the cost for it is a bound between the IO trait and the IO manager. In other words, callers of the IO trait methods must be stateful by keeping track of a cookie/token. Using this kind of API builds an implicit dependency between the trait and a registration method. To me, that's not the cleanest design, and the benefit we'd get for it is arguable (an O(1) lookup).

Ok, let's use (base, offset) tuple, otherwise we are just blocking by small technical points:)

src/lib.rs
```diff
@@ -20,15 +41,47 @@ pub enum IoAddress {
...
+        self.raw_value().partial_cmp(&other.raw_value())
+    }
+}
+
 /// Device IO trait.
 /// A device supporting memory based I/O should implement this trait, then
 /// register itself against the different IO type ranges it handles.
 /// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
 /// registered devices read or write method from this trait.
+/// The DeviceIo trait adopts the interior mutability pattern
```
alexandruag: Having interior mutability here is a thorny issue and needs a separate discussion :( Why do you favor this over the previous interface?

jiangliu: Some users of firecracker have mentioned that lacking support of multi-queue causes performance penalty, so we hope we could enable mutli-queue for virtio devices.
Moving lock into the device backend driver is the first step:)

sameo: > Having interior mutability here is a thorny issue and needs a separate discussion :( Why do you favor this over the previous interface?

@alexandruag Any further comments on this? I'm ok with this PR, but I'd like to get your opinion on it and see if you want to block it until we agree on the inner mutability part.

alexandruag: Hi, apologies for the delay. The problem with assuming interior mutability is many devices mutate their state during normal operation, but have to go out of their way to actually implement interior mutability. I don't know what the best approach is long term (hopefully it'll become clearer as we start putting things together), but one potential solution right now is to add another trait (something like `DeviceIoMut`) that defines the same methods as `DeviceIo` but requiring `&mut self` instead of `&self`, and then also add a blanket implementation such as `impl<T> DeviceIo for Mutex<T> where T: DeviceIoMut { ... }`, that allows a `Mutex<T>` to be used as a `DeviceIo` object automatically as long as the inner type implements `DeviceIoMut` (same goes for other standard wrappers that offer interior mutability like `RefCell`). Does that make any sense?

jiangliu: Great idea, this satisfies both side with an zero overhead abstraction.
How about merging the PR first and sending another PR for the DeviceIoMut proposal?

PR #13, opened 2019-11-13
Add device specific token to address different IO windows

liujing2:
Device can have several PIO or MMIO address spaces like PCI BARs.
Besides, not all devices have to know exactly the base IO address and
only need to track the offset address of each IO window. This patch adds `token`
to uniquely identify the different IO windows when IO VMExit, so that
device can use `token` and `offset` to address the data handling.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,14 +20,16 @@ pub enum IoAddress {
 }
 
 /// Device IO trait.
-/// A device supporting memory based I/O should implement this trait, then
-/// register itself against the different IO type ranges it handles.
-/// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
-/// registered devices read or write method from this trait.
+/// A device supporting memory based I/O should implement this trait. For
+/// device that has one or several IO (PIO or MMIO) address space, it
+/// registers itself against the different IO type ranges it handles with
+/// a unique token to distinguish different windows. The VMM will then
+/// dispatch IO (PIO or MMIO) VM exits by calling into the registered devices
+/// read or write method from this trait.
 pub trait DeviceIo: Send {
-    /// Read from the guest physical address `addr` to `data`.
-    fn read(&mut self, addr: IoAddress, data: &mut [u8]);
+    /// Read from `offset` of IO address space specified by `token` to `data`.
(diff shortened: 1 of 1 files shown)


liujing2:
@sameo This is a tiny PR for replacing full address parameter to token+offset parameters so that not every device has to record the base address but only need to know each IO window address token during registering. 

Please also take a look. @jiangliu

jiangliu:
> @sameo This is a tiny PR for replacing full address parameter to token+offset parameters so that not every device has to record the base address but only need to know each IO window address token during registering.
> 
> Please also take a look. @jiangliu

This helps to support PCI devices:)

src/lib.rs
```diff
@@ -20,14 +20,16 @@ pub enum IoAddress {
...
 
 /// Device IO trait.
-/// A device supporting memory based I/O should implement this trait, then
-/// register itself against the different IO type ranges it handles.
-/// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
-/// registered devices read or write method from this trait.
+/// A device supporting memory based I/O should implement this trait. For
+/// device that has one or several IO (PIO or MMIO) address space, it
+/// registers itself against the different IO type ranges it handles with
+/// a unique token to distinguish different windows. The VMM will then
```
sameo: > a unique token to distinguish different windows.

If anything, this token should naturally be the base address of that window. And yes, that base address would change when reprogramming PCI BARs, but so would the registration against the MMIO or IO bus. A token is well defined for PCI BARs, but it makes the API less simple and obvious.

jiangliu: We may use the base address as the token, but we may also have better choices for token.
For example, PCI bar index may be better than base address because we almost need to map the base address to bar index.
So token gives more flexibility to the driver implementation. And it would be easier to support firecracker by simply ignoring the token argument.

sameo: The problem with adding a token with VMM and device specific semantics, is that you have to use your `DeviceIo` instances with a specific device manager, one that can forward a cookie/token that was passed during device registration time. Your device IO trait becomes semantically bound to your device manager, which I feel is not the cleanest approach.
Adding a `base: IoAddress` argument would already be useful and help devices.

liujing2:
Close this to move else together into #18

PR #15, opened 2019-11-18
[WIP] Add emulation for Serial Console

andreeaflorescu:
This code is mostly based on CrosVM/Firecracker Serial implementation.

Still WIP.

Need to add/change:
- [ ] documentation
- [ ] comments with all the magic
- [ ] refine implementation of DeviceIo. IoAddress::PioAddress and IoAddress::MmmioAddress are making the implementation a bit weird.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,4 +6,6 @@ repository = "https://github.com/rust-vmm/vm-device"
 license = "Apache-2.0"
 
 [dependencies]
+log = "0.4"
 vm-memory = { git = "https://github.com/rust-vmm/vm-memory" }
+vmm-sys-util = ">0.1"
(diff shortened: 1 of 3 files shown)


jiangliu:
Will we include device drivers into the vm-device crate or create dedicated crate for device drivers?

andreeaflorescu:
We discussed about this at the rust-vmm meetup as well. If I remember correctly there were two approaches:
- each device or device type (i.e. legacy devices) in their own crate
- all devices in the same crate with a way of compiling them out if you don't need them

We reached a compromise in the sense that we can start having commonly used devices in vm-device. If the list of devices increases too much, we can then move out some devices to separate crates. I should send an email about this and get input from everyone.

andreeaflorescu:
Closing this PR from now. It will be moved to the legacy device crate once that exists.

PR #18, opened 2020-01-14
Remove vm-memory dependency and DeviceIo changing

liujing2:


src/lib.rs
```diff
@@ -25,9 +21,11 @@ pub enum IoAddress {
 /// The VMM will then dispatch IO (PIO or MMIO) VM exits by calling into the
 /// registered devices read or write method from this trait.
 pub trait DeviceIo: Send {
-    /// Read from the guest physical address `addr` to `data`.
-    fn read(&mut self, addr: IoAddress, data: &mut [u8]);
+    /// Read from guest physical address `base + offset` of the registered
```
andreeaflorescu: nit: We can write this as: "Read from the guest physical address `base` starting at `offset`. Result is placed in `data`." and then we don't need to specify what each argument is doing as it is more clear from the text. Same re-phrasing would be needed for write.

Unrelated to this change, but can someone remind me again why we don't return `data`? Having return values in function parameters is a bit of an anti-pattern.

sameo: > Unrelated to this change, but can someone remind me again why we don't return `data`? Having return values in function parameters is a bit of an anti-pattern.

It's more efficient, as it allows for a zero copy from the device into the KVM run IO/MMIO data structure.

PR #23, opened 2020-03-03
Yet another interrupt traits pull request

bonzini:
Introduce traits InterruptManager and Interrupt to manage interrupt sources for virtual devices, as well as a Vec wrapper that includes convenience `enable` and `disable` methods.

See also discussion in #21


src/interrupt/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+    ///
+    /// An interrupt notifier allows for external components and processes
+    /// to inject interrupts into a guest, by writing to the file returned
+    /// by this method.
+    fn notifier(&self) -> Option<&EventFd> {
+        None
+    }
+
+    /// Inject an interrupt from this interrupt source into the guest.
+    fn trigger(&self) -> Result<()>;
```
dwmw2: For level triggered interrupts, do we need a corresponding stop_triggering() function?

src/interrupt/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+pub trait MSIInterrupt: Interrupt {
```
dwmw2: At this level isn't "MSI" an implementation detail, and couldn't this be called `MaskableInterrupt`?

bonzini: Good idea.

src/interrupt/mod.rs
bonzini: I would rather avoid making the interrupt trait stateful. Instead, I would add an `eoi()` method instead. Level-triggered interrupts can check if the level is stil 1 and if so re-inject. The method would be invoked logically from the CPU, but it would also map nicely to KVM resamplefd semantics so there should be another `resample_notifier()` method.

sameo:
@jiangliu Seeing your SOB on this one, I assume you're ok with that implementation?

bonzini:
No, SoB only tracks the provenance of the code.

bonzini:
Anybody willing to review this or even try to integrate it in c-h? @sboeuf? :)

sboeuf:
@bonzini 
>Anybody willing to review this or even try to integrate it in c-h? @sboeuf? :)

If I have some free cycles next week I'll try that :)

bonzini:
Any news?

sboeuf:
I got pulled into migration stuff for CH, I don't have the bandwidth for now.

bonzini:
Well, even a review can help...

PR #29, opened 2020-09-23
Add simple Trigger trait for generating events

alexandruag:
This PR introduces the simple `Trigger` trait. I've started looking at the pending interrupt proposals and it seems this is a fundamental piece of functionality that we want to include anyway. Also, it can already help with abstracting away things such as using `EventFd`s to generate events/interrupts for `vm-superio` devices.


src/interrupt/mod.rs
```diff
@@ -0,0 +1,10 @@
+use std::result::Result;
```
aghecenco, line 1: Missing copyright.

alexandruag: fixed!

aghecenco (approved):

andreeaflorescu:
LGTM. This is the right step towards abstracting away the hypervisor from the emulation layer as well (i.e. looks like what we would need to fix: https://github.com/rust-vmm/vm-superio/issues/7).

src/interrupt/mod.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::result::Result;
+
+/// Abstraction for a simple, push-button like interrupt mechanism.
+pub trait Trigger {
+    /// Underlying type for the potential error conditions returned by `Self::trigger`.
+    type E;
```
JonathanWoollett-Light, line 9: It would be better and more typical to use `Error` for the generic error type (this is used https://doc.rust-lang.org/std/convert/trait.TryFrom.html for instance).

PR #31, opened 2020-10-26
IoManager: add Send + Sync constraint to PIO & MMIO buses

### LINKED ISSUE #30: Thread safe device management
The current implementation of `struct IoManager` (and, generally, the device buses) can't be shared across threads (== sent to the vCPUs) because the `Send` and `Sync` constraints aren't enforced anywhere.

aghecenco:
When sharing the device manager (`IoManager`) with the vCPUs, the devices on the buses need to be thread-safe. One way to do this is to explicitly enforce `Send + Sync` on `struct IoManager`'s buses; another would be to have `Send + Sync` as supertraits for `DevicePio` and `DeviceMmio`.

Fixes #30

### DIFF
--- a/src/device_manager.rs
+++ b/src/device_manager.rs
@@ -160,33 +160,33 @@ where
 #[derive(Default)]
 pub struct IoManager {
     // Range mapping for VM exit pio operations.
-    pio_bus: PioBus<Arc<dyn DevicePio>>,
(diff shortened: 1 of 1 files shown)


alexandruag (approved):
Looks good to me, and seems a bit more flexible than adding `Send` and `Sync` as supertraits. We can easily revert to the other approach if more arguments appear to support that direction.

andreeaflorescu (approved):

Issue #57, opened 2021-11-08
Define specialized type for the MMIO offset as well


andreeaflorescu:
We defined a type for the [PIO offset](https://github.com/rust-vmm/vm-device/blob/2c308e3e1968ca8119009b71e615086ba71e3bc5/src/lib.rs#L53).

Should we do the same for the MMIO one? Right now it is defined as an `u64`: https://github.com/rust-vmm/vm-device/blob/2c308e3e1968ca8119009b71e615086ba71e3bc5/src/lib.rs#L65

PR #58, opened 2021-11-09
Define types for address offsets

gsserge:
~This type alias is added for the symmetry with PioAddressValue.~

Add `MmioAddressOffset` and rename `PioAddressValue` to `PioAddressOffset` to make the naming more clear.

Closes https://github.com/rust-vmm/vm-device/issues/57

### DIFF
--- a/src/bus/address.rs
+++ b/src/bus/address.rs
@@ -30,17 +30,19 @@ pub trait BusAddress:
     fn checked_add(&self, value: Self::V) -> Option<Self>;
 }
 
+/// Represents a MMIO address offset.
+pub type MmioAddressOffset = u64;
+
 /// Represents a MMIO address.
 #[derive(Clone, Copy, Debug)]
-pub struct MmioAddress(pub u64);
+pub struct MmioAddress(pub MmioAddressOffset);
 
-/// This type defines the underlying value type for PIO addresses, which might be different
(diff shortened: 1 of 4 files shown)


andreeaflorescu (dismissed):

src/bus/address.rs
```diff
@@ -67,21 +70,21 @@ impl Ord for MmioAddress {
 impl Add<u64> for MmioAddress {
```
lauralt, line 70: Shouldn't this be `Add<MmioAddressValue>`?

jiangliu: Logically it's "address + offset", so do we need MmioOffset?

andreeaflorescu: The same thing happens for PIO Address. We were thinking about making the interface uniform for both PIO and MMIO. If we don't need the `MmioAddressValue` than we should remove the `PioAddressValue`. It's also a bit confusing (the difference between `Address` and `AddressValue`).

gsserge: Seems like everybody is confused about the ...Value names, so I think it's better to make things more clear and use `MmioAddressOffset` and `PioAddressOffset` instead.

andreeaflorescu (approved):

jiangliu (approved):

PR #62, opened 2021-11-30
Improve test coverage

gsserge:
Add new tests for device double registration, bus::check_access(),
and resources deregistration.

In addition, bus::register() does not require an additional check
that a new range is not present in the tree, because the range is
already checked for overlapping with any existing range.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+kcov_output/
--- a/src/bus/mod.rs
+++ b/src/bus/mod.rs
@@ -92,10 +92,6 @@ impl<A: BusAddress, D> Bus<A, D> {
             }
         }
 
-        if self.devices.contains_key(&range) {
-            return Err(Error::DeviceOverlap);
-        }
-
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+kcov_output/
```
andreeaflorescu, line 4: Can you add a PR in crate-template as well to improve the .gitignore there for future crates that we might add to rust-vmm?

src/bus/mod.rs
```diff
@@ -189,6 +185,9 @@ mod test {
             }
         }
 
+        // Detect double registration with the same range.
+        assert_eq!(bus.register(range, device), Err(Error::DeviceOverlap));
```
andreeaflorescu, line 189: Can we also check the overlap when the address start is not the same?

gsserge: This overlap is checked right below this one in the tests

lauralt (approved):

andreeaflorescu (approved):

Issue #28, opened 2020-09-12
Improve device manager, bus, and related documentation


alexandruag:
Improve the documentation of the public modules/structs/traits/functions related to the IO manager and bus abstractions.

PR #63, opened 2021-12-01
Improve documentation and examples

gsserge:
Closes https://github.com/rust-vmm/vm-device/issues/28

### DIFF
--- a/src/bus/range.rs
+++ b/src/bus/range.rs
@@ -41,6 +41,7 @@ impl<A: BusAddress> BusRange<A> {
         self.base
     }
 
+    /// Return the size of the range.
     pub fn size(&self) -> A::V {
         self.size
     }
@@ -81,8 +82,9 @@ impl<A: BusAddress> Ord for BusRange<A> {
     }
 }
 
-// Helper type aliases.
+/// Represents an MMIO bus range.
 pub type MmioRange = BusRange<MmioAddress>;
+/// Represents a PIO bus range.
 pub type PioRange = BusRange<PioAddress>;
 
 #[cfg(test)]
--- a/src/device_manager.rs
+++ b/src/device_manager.rs
@@ -3,12 +3,116 @@
 
 //! System level device management.
 //!
-//! [IoManager](struct.IoManager.html) is respondsible for managing
+//! [`IoManager`] is responsible for managing
 //! all devices of virtual machine, registering IO resources callback,
 //! deregistering devices and helping VM IO exit handling.
-//！VMM would be responsible for getting device resource request, ask
-//! vm_allocator to allocate the resources, ask vm_device to register the
-//! devices IO ranges, and finally set resources to virtual device.
+//! It defines two buses, one for PIO and one for MMIO, and provides default
+//! implementations of [`PioManager`] and [`MmioManager`].
+//!
+//! The VMM must first allocate unique resources (such as bus ranges), and then
+//! call into the vm-device interface to register the devices with their
+//! corresponding resources.
+//!
+//! # Examples
+//!
+//! Registering a new device can be done using the register methods of [`PioManager`]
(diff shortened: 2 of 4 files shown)


src/device_manager.rs
```diff
@@ -3,12 +3,116 @@
...
+//!
+//! Registering a new device can be done using the register methods of [`PioManager`]
+//! and [`MmioManager`] with an appropriate bus range
+//! ([`PioRange`](../bus/type.PioRange.html) or [`MmioRange`](../bus/type.MmioRange.html)).
+//! ```
+//! # use std::sync::Arc;
+//! # use vm_device::bus::{PioAddress, PioAddressOffset, PioRange};
+//! # use vm_device::bus::{MmioAddress, MmioAddressOffset, MmioRange};
+//! # use vm_device::device_manager::{IoManager, PioManager, MmioManager};
+//! # use vm_device::{DevicePio, DeviceMmio};
```
andreeaflorescu, line 26: nit: there are lots of imports compared to the number of lines of example code, maybe we can hide them from the example by using ` # ` in front of `use ...`.

andreeaflorescu (dismissed):

src/device_manager.rs
```diff
@@ -3,12 +3,116 @@
...
 //! deregistering devices and helping VM IO exit handling.
-//！VMM would be responsible for getting device resource request, ask
-//! vm_allocator to allocate the resources, ask vm_device to register the
-//! devices IO ranges, and finally set resources to virtual device.
+//! It defines two buses, one for PIO and one for MMIO, and provides default
+//! implementations of [`PioManager`] and [`MmioManager`].
+//!
+//! VMM would be responsible for getting device resource request, allocating the
+//! resources, asking `vm_device` to register the devices IO ranges, and finally
+//! setting resources to virtual device.
```
lauralt, line 14: I know this phrase wasn't added in this PR, but it sounds a bit strange, e.g. `asking vm_device something`; the `getting device resource request` is also confusing (for me at least). Can we somehow rephrase it please?

src/lib.rs
```diff
@@ -1,7 +1,52 @@
 // Copyright © 2019 Intel Corporation. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
-//! rust-vmm device model.
```
lauralt, line 4: I think we can also add the `#![deny(missing_docs)]` now. It also looks like [this public method](https://github.com/rust-vmm/vm-device/blob/main/src/bus/range.rs#L44) is not documented yet.

andreeaflorescu: Good catch!

src/device_manager.rs
gsserge: Please suggest how can we rephrase this.

andreeaflorescu: "The VMM must first allocate unique resources (such as bus ranges), and then call into the `vm-device` interface to register the devices with their corresponding resources."
WDYT?

lauralt: Sounds good to me, thanks.

src/bus/range.rs
```diff
@@ -81,8 +82,9 @@ impl<A: BusAddress> Ord for BusRange<A> {
     }
 }
 
-// Helper type aliases.
+/// Represents a MMIO bus range.
```
lauralt, line 85: nit:
```suggestion
/// Represents an MMIO bus range.
```

src/bus/range.rs
```diff
@@ -41,6 +41,7 @@ impl<A: BusAddress> BusRange<A> {
         self.base
     }
 
+    /// Return the number of attached devices.
```
lauralt, line 44: Isn't this returning the size of the range?

lauralt (approved):

andreeaflorescu (approved):

Issue #68, opened 2021-12-29
Requirements for interrupt abstractions in vm-device


alsrdn:
I'm opening this issue to start the discussion around adding interrupt abstractions to the vm-device crate.

# Motivation
Firecracker is working on adding support for PCI devices which naturally require support for MSI/MSI-X interrupts. CloudHypervisor already has PCI support implemented and has added support for MSI interrupts. Firecracker currently does not support MSI interrupts and uses a very simple model based on irqfds to implement interrupt delivery.

There is a large amount of device code that can be reused and shared across both CloudHypervisor and Firecracker that is not quite interoperable between VMMs today, one of the reasons being that they use different ways to work with interrupts.

# Requirements
The goal is to create some abstractions in vm-device that can model the interface for working with interrupts.
This would allow devices to have VMM-independent implementations that can be reused.

Devices should be able to:
* generate events that signal and interrupt the VM vCPUs
* configure interrupts based on a pre-defined specification
* enable/disable interrupt generation
* mask/unmask interrupts
* request new interrupts

The VMM should be able to:
* implement different interrupt mechanisms based on the machine specification
* manage interrupt sources
* deliver events generated by the devices to the vCPUs
* not need device specific knowledge or implementation (apart from standard configuration templates)

General requirements:
* interrupt events should not depend on the type of hypervisor used by the VM
* while new interrupt mechanisms rarely get added, the interfaces should allow for future extension and addition of new interrupt types
* vm-device can hold standard interfaces and templates for common interrupt types (e.g. ISA, INTx, MSI)

Even though the interface will allow VMMs to implement interrupt mechanisms independently, it may be a good idea to also have a reference implementation in vm-device for a common use-case (e.g. KVM in-kernel IRQChip).

# Prior work
There have been several PRs to add Interrupt management support to vm-device (https://github.com/rust-vmm/vm-device/pull/21 and https://github.com/rust-vmm/vm-device/pull/23).
There is also a different vm-device fork implementing these types of abstractions in the Cloud Hypervisor Github org (https://github.com/cloud-hypervisor/vm-device/pull/5) and an implementation that Cloud Hypervisor uses (https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/vm-device/src/interrupt/mod.rs).
[...]

PR #71, opened 2022-01-25
Interrupt abstractions

alsrdn:
This PR addresses requirements from #68 and adds interrupt abstractions to vm-device allowing devices to be developed independently from the vmm and to use a common interface for configuring and triggering interrupts.

This PR is a simplified version of https://github.com/rust-vmm/vm-device/pull/23 and https://github.com/rust-vmm/vm-device/pull/21 that allows for defining of custom behaviour in the VMM for different interrupt types and also interrupt groups and also allows for easy extension and creation of other interrupt types in the future by combining Interrupt super traits and configuration types.

I've included a small example on the usage of the traits.
Another example on how it might be used in Firecracker can be found in these draft commits: https://github.com/alsrdn/firecracker/commit/e1d6b7b5c72fd482e91fea60608d35bb365758cc and https://github.com/alsrdn/firecracker/commit/d59bd41bd8748bd3eb53623cbff8f4fe7f627198

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ address range, IRQ number, etc)
 
 ## Design
 
+### I/O
+
 The virtual device model is built around four traits, `DevicePio` and
 `MutDevicePio` for
 [Programmed I/O](https://en.wikipedia.org/wiki/Programmed_input%E2%80%93output)
@@ -42,8 +44,40 @@ that there’s a virtual device registered on the bus for the requested address,
 and finally sends the request to that device.
 ![vm-device](https://user-images.githubusercontent.com/241037/143853115-b1526028-6836-4845-a311-71cf989c60ef.png)
 
+### Interrupts
+
+Interrupt configuration is built around the `Interrupt` and `InterruptSourceGroup`
+traits. These traits allow device control code that is developed in separate
+crates from the VMM (which typically implements the interrupt mechanisms) to 
+configure interrupt delivery to the guest VM without having a dependency on the
+implementation of the interrupt mechanism.
+The `EdgeInterrupt` and `LevelInterrupt` are traits that provide interrupt
+assertion mechanisms and can be used by device implementations directly in order
+to inject interrupts in the guest VM.
+
+The `Interrupt` trait provides methods that are used by devices to enable and
+disable interrupt generation by registering the interrupt with the VMM. Access
+to additional interrupt properties is defined in new super-traits.
+`ConfigurableInterrupt` allows for devices to send or receive interrupt
+configuration parameters to/from the implementation inside the VMM. This is useful
+when devices need to specify custom data that the VMM will use when delivering the
+interrupt (e.g. MSI device id, PCI INTx pin etc).
+`MaskableInterrupt` is also defined as a super-trait for use with interrupts that
+can be masked/unmasked.
+
+An `InterruptSourceGroup` stores a collection of interrupts of the same type. It
+is the interface through which a device may request or release interrupts and
+perform group related actions like enabling or disabling all interrupts at once.
+Each device that generates interrupts can be assigned one or more
+`InterruptSourceGroup`s (depending on the types of interrupts it uses or logical
+grouping). The following diagram depicts the interaction between the components
+that use the interrupt interface:
+
+![vm-device-interrupts](https://user-images.githubusercontent.com/86006646/148783015-fea49a7c-cff8-4ec7-8766-00b0baed41c5.png)
+
 ## Usage
 
+### I/O
 A device is usually attached to a particular bus and thus needs to implement a
 trait of only one type. For example, serial port on x86 is a PIO device, while
 VirtIO devices use MMIO. It’s also possible for a device to implement both. Once
@@ -66,6 +100,44 @@ From now on the IoManager will be routing I/O requests for the registered
 address range to the device. The requests are dispatched by the client code, for
 example when handling VM exits, using `IoManager`'s methods `pio_read`,
 `pio_write`, `mmio_read` and `mmio_write`.
+
+### Interrupts
+
+To allow configuration of interrupts, a VMM must implement the `Interrupt` and
+`InterruptSourceGroup` traits for the interrupt mechanisms that the device
+requires. Implementation is machine or VMM specific and may depend on the types
+and number of IRQ chips that the machine has or interrupt delivery mechanisms
+(e.g. `EventFd`s).
+The device interrupt configuration code generally does not concern itself with
+the actual implementation of the interrupts and will be initialized with one or
+more `InterruptSourceGroup`s by the VMM.
+
+In order to allow devices to inject interrupt lines in the guest, a VMM must
+also implement either the `EdgeInterrupt` or `LevelInterrupt` traits, depending
+on the type of interrupt assertion mechanism that is supported. Devices use
+objects that have these traits to signal an interrupt to the guest.
+
+The device configuration code may define constraints for the types of interrupts
+that it needs by combining the supertraits defined into trait bounds (e.g. if it
+needs a `ConfigurableInterrupt` that can receive the `LegacyIrqConfig`
+configuration struct). MSI and legacy interrupt traits are added in this crate
+for ease of use. Configuration of these interrupt types are standardised.
+`MsiInterrupt` can also be used for MSI-X interrupts. These traits only define
+the configuration bounds for these interrupt types. `EdgeInterrupt` or
+`LevelInterrupt` will still need to be implemented in order to allow devices
+to use these interrupts. MSI interrupts are considered edge-triggered while
+`LegacyInterrupt` can be either edge-triggered (typically in the case of ISA
+interrupts) or level triggered (in the case of INTx interrupts).
+
+In order to have access to the underlying notification mechanisms used by the
+hypervisor, the device configuration may use the `AsRefTriggerNotifier` and the 
+`AsRefResampleNotifier` conversion traits and specify the `NotifierType` associated
+type. This type defines the interrupt delivery mechanism and is specific to the
+Hypervisor (e.g. KVM irqfd, Xen evtchn etc).
+
+One example of this requirement is the development of a VFIO device. Since VFIO
+can trigger a KVM irqfd directly, the VFIO device would need to get access to the
+underlying irqfd in order to register it with VFIO.
 
 ## Examples
 
@@ -119,3 +191,4 @@ This project is licensed under either of:
 
 - [Apache License](http://www.apache.org/licenses/LICENSE-2.0), Version 2.0
 - [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
+
--- a/src/interrupt/legacy.rs
+++ b/src/interrupt/legacy.rs
@@ -1,0 +1,39 @@
+// Copyright (C) 2021 Amazon.com, Inc. or its affiliates.
+// All Rights Reserved.
+
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Traits and Structs to manage legacy interrupt sources for devices.
+//!
+//! Legacy interrupt sources typically include pin based interrupt lines.
+
+use crate::interrupt::ConfigurableInterrupt;
+
+/// Definition for PCI INTx pins.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd)]
+pub enum IntXPin {
+    /// INTA
+    IntA = 0x1,
+    /// INTB
+    IntB = 0x2,
+    /// INTC
+    IntC = 0x3,
+    /// INTD
+    IntD = 0x4,
+}
+
+/// Standard configuration for Legacy interrupts.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+pub struct LegacyIrqConfig {
+    /// Input of the system interrupt controllers the device's interrupt pin is connected to.
+    /// Implemented by any device that makes use of an interrupt pin.
+    pub interrupt_line: Option<u32>,
+    /// Specifies which interrupt pin the device uses.
+    pub interrupt_pin: Option<IntXPin>,
+}
+
+/// Trait for defining properties of Legacy interrupts.
+pub trait LegacyInterrupt: ConfigurableInterrupt<Cfg = LegacyIrqConfig> {}
+
+/// Blanket implementation for Interrupts that use a LegacyIrqConfig.
+impl<T> LegacyInterrupt for T where T: ConfigurableInterrupt<Cfg = LegacyIrqConfig> {}
--- a/src/interrupt/mod.rs
+++ b/src/interrupt/mod.rs
@@ -1,0 +1,289 @@
+// Copyright (C) 2019-2020 Alibaba Cloud, Red Hat, Inc and Amazon.com, Inc. or its affiliates.
+// All Rights Reserved.
+
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Traits and Structs to manage interrupt sources for devices.
+//!
+//! In system programming, an interrupt is a signal to the processor emitted by hardware or
+//! software indicating an event that needs immediate attention. An interrupt alerts the processor
+//! to a high-priority condition requiring the interruption of the current code the processor is
+//! executing. The processor responds by suspending its current activities, saving its state, and
+//! executing a function called an interrupt handler (or an interrupt service routine, ISR) to deal
+//! with the event. This interruption is temporary, and, after the interrupt handler finishes,
+//! unless handling the interrupt has emitted a fatal error, the processor resumes normal
+//! activities.
+//!
+//! Hardware interrupts are used by devices to communicate that they require attention from the
+//! operating system, or a bare-metal program running on the CPU if there are no OSes. The act of
+//! initiating a hardware interrupt is referred to as an interrupt request (IRQ). Different devices
+//! are usually associated with different interrupts using a unique value associated with each
+//! interrupt. This makes it possible to know which hardware device caused which interrupts.
+//! These interrupt values are often called IRQ lines, or just interrupt lines.
+//!
+//! Nowadays, IRQ lines is not the only mechanism to deliver device interrupts to processors.
+//! MSI [(Message Signaled Interrupt)](https://en.wikipedia.org/wiki/Message_Signaled_Interrupts)
+//! is another commonly used alternative in-band method of signaling an interrupt, using special
+//! in-band messages to replace traditional out-of-band assertion of dedicated interrupt lines.
+//! While more complex to implement in a device, message signaled interrupts have some significant
+//! advantages over pin-based out-of-band interrupt signaling. Message signaled interrupts are
+//! supported in PCI bus since its version 2.2, and in later available PCI Express bus. Some non-PCI
+//! architectures also use message signaled interrupts.
+//!
+//! While IRQ is a term commonly used by Operating Systems when dealing with hardware
+//! interrupts, the IRQ numbers managed by OSes are independent of the ones managed by VMM.
+//! For simplicity sake, the term `Interrupt Source` is used instead of IRQ to represent both pin-based
+//! interrupts and MSI interrupts.
+
+pub mod legacy;
+pub mod msi;
+
+use std::fmt::{self, Display};
+use std::ops::Deref;
+use std::sync::Arc;
+
+/// Errors associated with handling interrupts
+#[derive(Debug)]
+pub enum Error {
+    /// Operation not supported for this interrupt.
+    OperationNotSupported,
+
+    /// The specified configuration is not valid.
+    InvalidConfiguration,
+
+    /// The interrupt state was not changed.
+    InterruptNotChanged,
+
+    /// The interrupt could not be triggered, asserted or de-asserted.
+    InterruptNotTriggered,
+
+    /// An error occured during interrupt allocation.
+    InterruptAllocationError,
+
+    /// An error occured during interrupt release.
+    InterruptFreeError,
+}
+
+impl std::error::Error for Error {}
+
+/// Reuse std::io::Result to simplify interoperability among crates.
+pub type Result<T> = std::result::Result<T, Error>;
+
+impl Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Interrupt error: ")?;
+        match self {
+            Error::OperationNotSupported => write!(f, "operation not supported"),
+            Error::InvalidConfiguration => write!(f, "invalid configuration"),
+            Error::InterruptNotChanged => write!(f, "the interrupt state could not be changed"),
+            Error::InterruptNotTriggered => write!(f, "the interrupt could not be triggered"),
+            Error::InterruptAllocationError => write!(f, "the interrupt could not be allocated"),
+            Error::InterruptFreeError => write!(f, "the interrupt could not be released"),
+        }
+    }
+}
+
+/// Trait used by interrupt producers to emulate an edge triggered interrupt.
+///
+/// This trait allows for a device to signal an interrupt event to the guest VM.
+/// These events are sampled by the interrupt controller synchronously when `trigger()` is called.
+///
+/// Edge triggered interrupts cannot be shared.
+pub trait EdgeInterrupt {
+    /// Signal an interrupt to the guest VM.
+    fn trigger(&self) -> Result<()>;
+}
+
+/// Trait used by interrupt producers to emulate a level triggered interrupt.
+///
+/// This trait allows for a device to assert an interrupt line as long as it needs service
+/// from the CPU.
+/// A level triggered interrupt is held asserted until the device clears the interrupt signal.
+///
+/// Assertion of the interrupt signal starts when `assert()` is called by the device and ends when
+/// `clear()` is called by the device.
+/// The object implementing this trait must hold the assertion state internally and should return
+/// from `assert()` and `clear()` once the state is changed.
+pub trait LevelInterrupt {
+    /// Assert the interrupt line to signal an interrupt to the guest VM.
+    /// This method sets the interrupt in an asserted state.
+    fn assert(&self) -> Result<()>;
+
+    /// Deassert the interrupt line to signal that the device no longer requires service.
+    fn clear(&self) -> Result<()>;
+}
+
+/// Trait that allows access to a device interrupt status.
+///
+/// A device will implement this trait if it wants to allow other components to check its
+/// interrupt status.
+/// This allows implementation of auto-triggered shared level interrupts that poll the interrupt
+/// state from the device in order to re-trigger the interrupt when resampled.
+pub trait InterruptStatusChecker {
+    /// Check if the device requires service.
+    /// Returns `true` if the device has not deasserted the interrupt line and still
+    /// requires service.
+    fn is_active(&self) -> bool;
+}
+
+/// Trait used by interrupt controllers to configure interrupts.
+///
+/// An object having the `Interrupt` trait is shared between the VMM (which typically implements
+/// the interrupt mechanisms) and interrupt control components.
+/// It offers a control interface through the `enable()` and `disable()` methods that allow an
+/// interrupt to be registered with the interrupt controllers or mechanisms in the VMM.
+///
+/// Objects implementing this trait are required to have internal mutability.
+pub trait Interrupt {
+    /// Enable generation of interrupts on this line.
+    fn enable(&self) -> Result<()> {
+        Err(Error::OperationNotSupported)
+    }
+
+    /// Disable generation of interrupts on this line.
+    fn disable(&self) -> Result<()> {
+        Err(Error::OperationNotSupported)
+    }
+}
+
+/// Trait that allows interrupt controllers to configure interrupt parameters.
+///
+/// This enhances the control plane interface of the `Interrupt` by allowing a device to configure
+/// the behavior of the interrupt.
+///
+/// Objects implementing this trait are required to have internal mutability.
+pub trait ConfigurableInterrupt: Interrupt {
+    /// Type describing the configuration spec of the interrupt.
+    type Cfg;
+
+    /// Update configuration of the interrupt.
+    fn update(&self, config: &Self::Cfg) -> Result<()>;
+
+    /// Returns the current configuration of the interrupt.
+    fn get_config(&self) -> Result<Self::Cfg>;
+}
+
+/// Trait for interrupts that can be masked or unmasked.
+///
+/// Objects implementing this trait are required to have internal mutability.
+pub trait MaskableInterrupt: Interrupt {
+    /// Mask the interrupt.  Masked interrupts are remembered but
+    /// not delivered.
+    fn mask(&self) -> Result<()>;
+
+    /// Unmask the interrupt, delivering it if it was pending.
+    fn unmask(&self) -> Result<()>;
+}
+
+/// Trait for interrupts that can be auto-retriggered when resampled.
+///
+/// In some implementations of shared level-triggered interrupts the interrupt can be resampled
+/// as a result of different events (e.g. an EOI) before a device explicitly deasserts the
+/// interrupt. If the device still requires service, the interrupt should be reasserted.
+///
+/// This trait allows implementation of the interrupt mechanism described above.
+/// It requires that the user of this trait configures an `InterruptStatusChecker`.
+/// When the interrupt is resampled, the state of the device will be checked by the implementation
+/// of this trait. If the device still requires service, the interrupt is reasserted.
+///
+/// An example of such mechanism is provided by KVM when using KVM_IRQFD with
+/// KVM_CAP_IRQFD_RESAMPLE.
+///
+/// Objects implementing this trait are required to have internal mutability.
+pub trait AutoRetriggerInterrupt: Interrupt {
+    /// Set the `InterruptStatusChecker` object through which the interrupt can poll the device
+    /// interrupt status.
+    fn set_status_checker(&self, status_checker: Arc<dyn InterruptStatusChecker>) -> Result<()>;
+}
+
+/// Trait that provides access to the underlying trigger notification object used by the hypervisor.
+///
+/// The type of the underlying notification mechanism used by the interrupt is defined by the
+/// `NotifierType` associated type.
+/// This enables some use cases where the device may want to bypass the VMM completely or when the
+/// device crate acts only as a control plane and the actual emulation is implemented in some other
+/// component that understands the underlying mechanism.
+///
+/// The usage of the resulted notifier object is speciffic to the hypervisor but the semantics of
+/// the object returned by the `trigger_notifier()` method should follow the semantics from
+/// `EdgeInterrupt::trigger()` or `LevelInterrupt::assert()` (e.g. when the user changes the state
+/// of the notifier object, an interrupt is queued for the guest).
+///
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -9,3 +9,7 @@ repository = "https://github.com/rust-vmm/vm-device"
 license = "Apache-2.0 OR BSD-3-Clause"
 
 [dependencies]
+vmm-sys-util = { version = "0.8.0" }
```
gsserge, line 12: Unless more parameters are used, `vmm-sys-util = "0.8.0"` is simpler. The same goes for the `libc` dev dependency below.

README.md
```diff
@@ -42,8 +44,38 @@ that there’s a virtual device registered on the bus for the requested address,
 and finally sends the request to that device.
 ![vm-device](https://user-images.githubusercontent.com/241037/143853115-b1526028-6836-4845-a311-71cf989c60ef.png)
 
+### Interrupts
+
+Interrupt delivery and configuration is build around the `Interrupt` and
```
gsserge, line 49:
```suggestion
Interrupt delivery and configuration is built around the `Interrupt` and
```

src/interrupt/legacy.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Copy, Clone, Debug, Default)]
+pub struct LegacyIrqConfig {
+    /// Input of the system interrupt controllers the device's interrupt pin is connected to.
+    /// Implemented by any device that makes use of an interrupt pin.
+    pub interrupt_line: Option<u32>,
+    /// Specifies which interrupt pin the device uses.
+    pub interrupt_pin: Option<IntXPin>,
+}
+
+/// Supertrait for defining properties of Legacy interrupts.
```
gsserge, line 35: `LegacyInterrupt ` is not a supertrait, it's a subtrait. The readme correctly states that `ConfigurableInterrupt` and `MaskableInterrupt` are supertraits. I would not focus on this here and just call it a trait.

src/interrupt/msi.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub high_addr: u32,
+    /// Low address to delivery message signaled interrupt.
+    pub low_addr: u32,
+    /// Data to write to delivery message signaled interrupt.
+    pub data: u32,
+    /// Unique ID of the device to delivery message signaled interrupt.
+    pub devid: u32,
+}
+
+/// Supertrait for defining properties of MSI interrupts.
```
gsserge, line 28: Same as `LegacyInterrupt`, `MsiInterrupt` is not a supertrait.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use std::fmt::{self, Display};
+use std::io;
+use std::ops::Deref;
+
+use crate::interrupt::trigger::Trigger;
+
+/// Errors associated with handling interrupts
+#[derive(Debug)]
+pub enum Error {
```
gsserge, line 57: It does not cost as anything to add an empty `impl std::error::Error for Error {}`, allowing any client code more flexible error handling if needed.

src/interrupt/trigger/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+/// This helps in abstracting away how events/interrupts are generated when
+/// working with the emulated devices.
+///
+/// The user has to provide a `Trigger` object to the device's constructor when
+/// initializing that device. The generic type `T: Trigger` is used in the
+/// device's structure definition to mark the fact that the events notification
+/// mechanism is done via the Trigger interface.
+pub trait Trigger {
+    /// Underlying type for the potential error conditions returned by `Self::trigger`.
+    type E: std::fmt::Debug;
```
gsserge, line 21: Why the `Debug` trait bound? If we want to communicate that `E` is an error, then the bound should be `std::error::Error`, unless there are good reasons not to. Alternatively we could drop the bound altogether and just leave the associated type as `type E;`.

src/interrupt/legacy.rs
alsrdn: Yup, you're right. Best to leave it as "Trait".

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Error {
+        Error::IOError(e)
+    }
+}
+
+/// Trait for interrupt producers.
+pub trait Interrupt {
```
gsserge, line 93: A few top-level comments about this trait:

1. I think it would be nice to mention in the doc comment that interior mutability is required by design (all methods use `&self`).
2. The default method implementations look a bit arbitrary. For example, why `acknowledge()` is a no-op,`enable()` is required, while `disable()` returns an error by default.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// The type of Trigger object used for notifications by this interrupt.
+    type TriggerType: Trigger;
+
+    /// Inject an interrupt from this interrupt source into the guest.
+    fn trigger(&self) -> std::result::Result<(), <Self::TriggerType as Trigger>::E>;
+
+    /// Returns an interrupt notifier from this interrupt.
+    ///
+    /// An interrupt notifier allows for external components and processes
+    /// to inject interrupts into a guest, by writing to the file returned
```
gsserge, line 103: Isn't `notifier()` also used for the vmm code to receive the interrupt notifications? That's how it's implemented in the example.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// by this method.
+    fn notifier(&self) -> Option<Self::TriggerType> {
```
gsserge, line 105: Why the return type is optional? Is there a possible scenario where a notifier is not present?
Same question about `ack_notifier()`.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+/// Trait to manage a group of interrupt sources for a device.
+///
+/// A device may use an InterruptSourceGroup to manage multiple interrupts of the same type.
+/// The group allows a device to request and release interrupts and perform actions on the
+/// whole collection of interrupts like enable and disable for cases where enabling or disabling
+/// a single interrupt in the group does not make sense. For example, PCI MSI interrupts must be
+/// enabled as a group.
+pub trait InterruptSourceGroup: Send {
+    /// Type of the interrupts contained in this group.
+    type InterruptType: Interrupt;
```
gsserge, line 162: Is there a real need for this associated type? Couldn't we get away with only `InterruptWrapper`?

src/interrupt/trigger/mod.rs
alsrdn: It's actually the definition moved from here: https://github.com/rust-vmm/vm-superio/blob/14906a0e5363557c9c97632b03f986135c4f5d3b/crates/vm-superio/src/lib.rs#L47

It actually makes more sense to bind it to `std::error::Error` though users of this trait from vm-superio might need to change their implementations. But I guess that maybe it's best to change it now rather than later.

andreeaflorescu: Can we instead merge this PR: https://github.com/rust-vmm/vm-device/pull/29/ and then rebase this one on top of it? We can do the adjustments in vm-superio afterwards as right now vm-superio defines its own trait anyway.

alsrdn: Actually I've imported the commit from #29 an rebased it on top of main. It is now part of this PR.

src/interrupt/mod.rs
alsrdn: I think we need it so that we can have a bound for the Deref target.

src/interrupt/trigger/mod.rs
lauralt: Can you also add Alex's signed-off line to the commit message? We are trying to keep the original author of the commits that we cherry-pick. Example [here](https://github.com/rust-vmm/vm-superio/commit/7f3cb2b6a3f42b4c5c1faf414e8bd35030ab4f9b).

src/interrupt/mod.rs
alsrdn: It can work both ways since the notifier contains 2 ends, one for the component that triggers the notification and one for the component being notified. Indeed the example shows only the part that is notified. Here's an example where the notifier is used by the device directly: https://github.com/alsrdn/firecracker/blob/d59bd41bd8748bd3eb53623cbff8f4fe7f627198/src/devices/src/virtio/net/device.rs#L768

src/interrupt/trigger/eventfd.rs
```diff
@@ -0,0 +1,59 @@
...
+
+//! Implementation for a Trigger based on EventFd.
+
+use std::io;
+use std::ops::Deref;
+use vmm_sys_util::eventfd::EventFd;
+
+use crate::interrupt::trigger::Trigger;
+
+/// Trait used for creating an EventFd from an object.
```
andreeaflorescu, line 14: Why is this needed? Intuitively you should be able to use the deref implementation to get a clone of the inner `EventFd`. The name of this trait is not appropriate for the implementation because `as` suggests a conversion from one type to another, but the implementation is actually creating a clone.

Here's an example:
```rust
let ev_trigger = EventFdTrigger::new(EventFd::new(0).unwrap());
let inner: &EventFd = ev_trigger.deref();
```

examples/interrupts/device.rs
```diff
@@ -0,0 +1,52 @@
+use std::sync::{Arc, Mutex};
```
andreeaflorescu, line 1: Do you mind dropping this commit from the initial PR? It is not required, and it might be better to first focus on the implementation, and then add the examples only once the interfaces are stabilized.

gsserge: I actually found the example to be useful for understanding the whole picture.

andreeaflorescu: Yes, I agree they're useful for understanding how the abstractions work, but I'd like to avoid getting too many comments on a single PR. The examples can be linked in the PR (as the usage in Firecracker is), and it would also go in line with the [recommendation of keeping the PRs under 250 lines](https://hugooodias.medium.com/the-anatomy-of-a-perfect-pull-request-567382bb6067#:~:text=Recap-,Pull%20request%20size,of%20250%20lines%20of%20change.)). We can also just ignore them in the first iterations of the PR, and just review them after the interfaces are stabilized.

src/interrupt/mod.rs
gsserge: Thanks for the example. What I meant was that the receiver part is quite important and should be mentioned in the doc comment. 

At least this is how I understand the high-level design. An `Interrupt` object is shared between device and vmm (via the `InterruptSourceGroup`, but the group is merely a collection). The object has two main methods:
- `trigger()` is used by the device to trigger the interrupt and notify vmm
- `notifier()` is used by the vmm to receive the triggered interrupt

examples/interrupts/device.rs
gsserge: I see what you mean. I think we can ignore the example in the first iterations to focus on the interfaces, and address them later on. @alsrdn would that be ok with you?

src/interrupt/mod.rs
alsrdn: Yes, an `Interrupt` object is shared by between the device and the vmm. It performs 2 main goals:
- interrupt control plane (through the `enable`, `disable` methods that extend to `update`, `mask`, `unmask` for `MaskableInterrupt` and `ConfigurableInterrupt`)
- interrupt data plane through which an event/notification can be delivered between the device and the vmm (`trigger` and `acknowledge`).

Apart from the vmm and the device crate, we have the Hypervisor which is the one responsible for actually delivering the notification to the guest. The Hypervisor typically has a well defined interface for sending notifications (KVM uses irqfds, Xen uses evtchn etc). an `Interrupt` wraps those mechanisms in a common interface and allows them to be configured. When a device uses `trigger`, the `Interrupt` will use this underlying mechanism to signal the Hypervisor.

This interface alone works well but, there are some use cases where you need access to the underlying mechanism in the device. This is true for any device crate that acts more as a control plane and does the emulation somewhere else. A notable example is VFIO that allows us to register the irqfd so that interrupts follow a fast path that doesn't require going through the VMM.
This is where the `notifier` is needed because it provides a way to access the underlying mechanism for delivering the event to the Hypervisor.

While writing this explanation it occurred to me that the `Trigger` trait only complicates the interface without bringing significant value over what already exists in `Interrupt`. My initial reasoning was that it would make for an easier transition between the crates that already use `Trigger`.
So, I would actually drop the `Trigger` completely and just have an associated `NotifierType` value which is unbounded. @gsserge what do you think?

examples/interrupts/device.rs
alsrdn: yup, no problem.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    OperationNotSupported,
+
+    /// The specified configuration is not valid.
+    InvalidConfiguration,
+
+    /// The interrupt was not enabled.
+    InterruptNotEnabled,
+
+    /// Generic IO error,
+    IOError(io::Error),
```
gsserge, line 68: (this was discovered by @alexandruag, I'm putting it here so that it does not get lost)

`io::Error` is probably too restrictive and also a bit awkward to enforce on the client code. I think we should either make this Error enum generic (basically moving `Trigger::E` into the enum's type parameter), or have `Generic(Box<dyn std::error::Error>)` as a variant to hold the (boxed) error coming from the client code.

Cargo.toml
alsrdn: Removed the dependency since EventFdTrigger was removed.

src/interrupt/legacy.rs
alsrdn: Corrected.

src/interrupt/msi.rs
alsrdn: Corrected.

src/interrupt/mod.rs
alsrdn: Added empty implementation for Error.

src/interrupt/mod.rs
alsrdn: Added comment calling out requirement for interior mutability.
Added default methods that return `OperationNotSupported` for all methods except `trigger()` for more consistency.

src/interrupt/mod.rs
alsrdn: I've removed the `Trigger` completely and added some more details in the documentation.

src/interrupt/mod.rs
alsrdn: Due to the associated type being present, there will always be a notifier present. But the implementor might decide not to make it available to the device.

src/interrupt/legacy.rs
```diff
@@ -0,0 +1,39 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+//! Traits and Structs to manage legacy interrupt sources for devices.
+//!
+//! Legacy interrupt sources typically include pin based interrupt lines.
+
+use crate::interrupt::ConfigurableInterrupt;
+
+/// Definition for PCI INTx pins.
+#[derive(Copy, Clone, Debug)]
```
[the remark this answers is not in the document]
alsrdn: Added common traits where applicable.

src/interrupt/trigger/eventfd.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl EventFdTrigger {
+    /// Try to clone this EventFdTrigger into a new object.
+    /// Returns an error if the underlying EventFd cannot be cloned.
+    pub fn try_clone(&self) -> io::Result<Self> {
+        Ok(EventFdTrigger((**self).try_clone()?))
+    }
+
+    /// Construct a new EventFdTrigger using the EventFd passed as parameter.
+    pub fn new(evt: EventFd) -> Self {
```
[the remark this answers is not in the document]
alsrdn: I've removed `EventFdTrigger` along with `Trigger` based on https://github.com/rust-vmm/vm-device/pull/71#discussion_r793388445.

src/interrupt/trigger/eventfd.rs
alsrdn: I've removed EventFdTrigger along with Trigger based on #71 (comment).

examples/interrupts/device.rs
alsrdn: Removed the example from this PR. Will add it later.

src/interrupt/mod.rs
alsrdn: Since we don't use `Trigger` anymore I removed `IOError` completely and added `InterruptNotTriggered`.

gsserge:
Please adjust test coverage to make CI happy, otherwise LGTM.

gsserge:
@jiangliu @studychao could you please also take a look at the PR, thanks!

src/interrupt/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+        None
+    }
+
+    /// Enable generation of interrupts from this interrupt source.
+    fn enable(&self) -> Result<()> {
+        Err(Error::OperationNotSupported)
+    }
+
+    /// Disable generation of interrupts from this interrupt source.
+    fn disable(&self) -> Result<()> {
```
blitz, line 142: This trait mixes up multiple aspects of an interrupt:

- The part that is seen by a device (trigger)
- A super-trait of Interrupt where the underlying notifier (eventfd) can be extracted (notifier)
- The part that is operated by interrupt controllers (enable, disable, ack?)

These should be separate traits to simplify implementation and re-usability.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Returns an interrupt notifier from this interrupt.
+    ///
+    /// An interrupt notifier allows for external components and processes
+    /// to inject interrupts into a guest through a different interface other
+    /// than `trigger`.
+    fn notifier(&self) -> Option<Self::NotifierType> {
+        None
+    }
+
+    /// Called back when the CPU acknowledges the interrupt.
```
blitz, line 121: What does this mean? The CPU does not acknowledge interrupts.

Is this called when the CPU actually receives the interrupt or when it is logged into the LAPIC or when the guest does an EOI?

src/interrupt/mod.rs
alsrdn: Yes, I think separating the notifier related functionality would improve the interface at least for the cases where the notifiers are not available.
For enable/disable I think moving them to another trait has diminishing returns. It's true that those methods are related to interrupt control but that can come either from VMM code or from a device crate code (e.g. a PCI crate which handles the parsing of MSI configuration).

src/interrupt/mod.rs
alsrdn: Need to update the doc a bit here. It's actually an EOI callback and it would be used for shared level-sensitive interrupts in order to inject the interrupt if the device still requires it (e.g. for INTx interrupts). It's equivalent to the `ack_notifier()` the difference being that the latter would enable the resampled semnatics with external components like vfio.

src/interrupt/mod.rs
alsrdn: I've moved the notifier methods to a different trait. Please take a look.

gsserge (approved):

src/interrupt/mod.rs
blitz: For level-triggered interrupts the `trigger` method seems not to be the right abstraction to me. In this case, the device would need a mechanism to assert/de-assert the interrupt line for correct emulation. Optimizations to avoid assert/deassert messages when the interrupt cannot be injected anyway would be a second step.

In any case, the flow for handling level triggered interrupts should be documented. I wouldn't have seen how this is supposed to work from the code alone.

src/interrupt/mod.rs
blitz: > For enable/disable I think moving them to another trait has diminishing returns. It's true that those methods are related to interrupt control but that can come either from VMM code or from a device crate code (e.g. a PCI crate which handles the parsing of MSI configuration).

I see this from the point of what I have to implement for example to unit test a device that toggles interrupts. A simple device would take an interrupt pin as an input, which can be anything that implements the `trigger` method. When I use a mock of this for unit testing with the current trait, I have to implement lots of unrelated things that the device should not even be able to call at all.

So in my mind it's very important to separate the two sides: One set of functions that the device is supposed to use. One set of functions that the interrupt controller is supposed to use. These should be two traits that also have no relation to each other. That you can implement them on the same struct is an implementation detail.

Or to put it a different way: The device should only be able to use interfaces that it is actually supposed to be using. This prevents people who may not fully understand the interrupt logic from using the interfaces the wrong way, which is a big win.

blitz: Maybe as clarification: We are currently using rust-vmm code for parts of our virtualization stack at Cyberus Technology internally. We are interested in making it more generic and thus a bit more friendly to hypervisors that don't necessarily work like KVM. :)

This is also a reason why I'm a bit concerned about mixing up concerns in trait implementations and keeping it simple, straightforward and without guess work for the implementor of traits, because this has already caused some churn for us when we onboarded with vm-memory.

alsrdn: Gotcha. I think it's good to have the separation for the use cases you mentioned. My simplifying assumption was that most implementations would mix both pure emulation functionality and some interrupt control code so it would be more easier to pass Interrupt objects that have more optional functionality in one larger interface.
But I'm definitely not against splitting them up. Here's what I propose (also merging the discussion about level-interrupts below):

- Interrupt traits **used by device emulation** to generate notifications:
```rust
pub trait EdgeInterrupt {
    fn trigger(&self) -> Result<()>;
}

pub trait LevelInterrupt {
    fn assert(&self) -> Result<()>;
    fn de_assert(&self) -> Result<()>;
}

/// Trait for interrupts that can be masked or unmasked.
/// Can be used with either Level or Edge triggered interrupts.
pub trait MaskableInterrupt {
    fn mask(&self) -> Result<()>;

    fn unmask(&self) -> Result<()>;
}
```

- Interrupt control traits used by interrupt controllers and interrupt configuration code:
```rust
pub trait Interrupt {
    /// Enable generation of interrupts from this interrupt source.
    fn enable(&self) -> Result<()>;

    /// Disable generation of interrupts from this interrupt source.
    fn disable(&self) -> Result<()>;
}

pub trait ConfigurableInterrupt: Interrupt {
    /// Type describing the configuration spec of the interrupt.
    type Cfg;

    /// Update configuration of the interrupt.
    fn update(&self, config: &Self::Cfg) -> Result<()>;

    /// Returns the current configuration of the interrupt.
    fn get_config(&self) -> Result<Self::Cfg>;
}

/// Trait implemented by devices through which a resampled interrupt that was cleared
/// can determine if the device still requires service
pub trait InterruptStatusChecker {
    fn is_active(&self) -> bool;
}

/// Trait for Interrupts that can be automatically retriggered when resampled
/// Uses InterruptStatusChecker
pub trait AutoRetriggerInterrupt: Interrupt {
    fn set_status_checker(&self, Arc<dyn InterruptStatusChecker>) -> Result<()>;
}

pub trait InterruptSourceGroup: Send {
    /// [...]
}
```

- Conversion traits to get the underlying notifiers (for the irqfd use-case)
```rust
pub trait AsRefTriggerNotifier {
    type NotifierType;
    fn trigger_notifier(&self) -> &Self::NotifierType;
}

pub trait AsRefResampleNotifier {
    type NotifierType;
    fn resample_notifier(&self) -> &Self::NotifierType;
}
```

@blitz WDYT?

gsserge: This suggestion looks good to me. The only think I'd like to change is the `de_assert()` method. Maybe `unassert()` or `clear()`?

alsrdn: I've updated the PR with the changes discussed here. @gsserge @blitz PTAL

gsserge (approved):

src/interrupt/msi.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use crate::interrupt::{ConfigurableInterrupt, MaskableInterrupt};
+
+/// Configuration data for MSI/MSI-X interrupts.
+///
+/// On x86 platforms, these interrupts are vectors delivered directly to the LAPIC.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+pub struct MsiIrqConfig {
+    /// High address to delivery message signaled interrupt.
+    pub high_addr: u32,
```
blitz, line 19: Why not `pub addr: u64`?

blitz (commented):
Everything looks documented, but I still don't feel like I understand the usecases that drive the complexity.

For devices that want to trigger MSIs, how does the interface look for them. Is there an example device somewhere?

src/interrupt/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn resample_notifier(&self) -> &Self::NotifierType;
+}
+
+/// Trait to manage a group of interrupt sources for a device.
+///
+/// A device may use an InterruptSourceGroup to manage multiple interrupts of the same type.
+/// The group allows a device to request and release interrupts and perform actions on the
+/// whole collection of interrupts like enable and disable for cases where enabling or disabling
+/// a single interrupt in the group does not make sense. For example, PCI MSI interrupts must be
+/// enabled as a group.
```
blitz, line 260: Is this a KVM limitation or something that comes as a requirement from PCI passthrough?

src/interrupt/msi.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub data: u32,
+    /// Unique ID of the device to delivery message signaled interrupt.
+    pub devid: u32,
+}
+
+/// Trait for defining properties of MSI interrupts.
+pub trait MsiInterrupt: ConfigurableInterrupt<Cfg = MsiIrqConfig> + MaskableInterrupt {}
+
+/// Blanket implementation for Interrupts that use a MsiIrqConfig.
+impl<T> MsiInterrupt for T where T: ConfigurableInterrupt<Cfg = MsiIrqConfig> + MaskableInterrupt {}
```
blitz, line 32: I have really no idea how this would be used by a device. Is there an example device to look at somewhere?

src/interrupt/msi.rs
alsrdn: This follows the PCI spec. See sections 6.8.1.4, 6.8.1.5 from here https://www.cl.cam.ac.uk/~djm202/pdf/specifications/pci/PCI_LB3.0_CB-2-6-04.pdf

alsrdn:
> Everything looks documented, but I still don't feel like I understand the usecases that drive the complexity.
> 
> For devices that want to trigger MSIs, how does the interface look for them. Is there an example device somewhere?

I currently removed the example to keep this PR more focused following the discussion here https://github.com/rust-vmm/vm-device/pull/71#discussion_r792849521.
In order to trigger MSIs in a device emulation, configuration code should allocate, configure and enable the interrupts and pass an object with the EdgeInterrupt trait to the emulation code that will use `trigger()` to trigger the interrupt.
Here's an example of legacy interrupts in firecracker: https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/device_manager/legacy.rs#L84 (it's outdated right now since it was using the old traits).

src/interrupt/mod.rs
alsrdn: MSI or MSI-x is enabled by setting the enable bit in the message control struct. This refers to the whole set of MSI interrupts. You can mask each individual interrupt but you can't enable/disable them individually.

src/interrupt/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Return the index-th interrupt in the group, or `None` if the index is out
+    /// of bounds.
+    fn get(&self, index: usize) -> Option<Self::InterruptWrapper>;
+
+    /// Request new interrupts within this group.
+    fn allocate_interrupts(&mut self, size: usize) -> Result<()>;
+
+    /// Release all interrupts within this group.
+    fn free_interrupts(&mut self) -> Result<()>;
```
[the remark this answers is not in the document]
alsrdn: `allocate_interrupts` and `free_interrupts` should be called from interrupt setup code to reserve/release interrupts.
Typically you would return an error when `enabled` is called on an empty group in the trait implementation.

src/interrupt/msi.rs
alsrdn: This is used for convenience. It becomes tedious to pass around a long list of trait bounds for well known interrupt types.

src/interrupt/mod.rs
blitz: Is there a functional difference between masked and disabled MSI-X vectors from the interrupt subsystems perspective?

src/interrupt/mod.rs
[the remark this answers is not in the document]
blitz: > should be called from interrupt setup code

What setup code are you referring to here? The whole "flow" of this API is not really clear to me yet.

blitz:
> > Everything looks documented, but I still don't feel like I understand the usecases that drive the complexity.
> > For devices that want to trigger MSIs, how does the interface look for them. Is there an example device somewhere?
> 
> I currently removed the example to keep this PR more focused following the discussion here [#71 (comment)](https://github.com/rust-vmm/vm-device/pull/71#discussion_r792849521). In order to trigger MSIs in a device emulation, configuration code should allocate, configure and enable the interrupts and pass an object with the EdgeInterrupt trait to the emulation code that will use `trigger()` to trigger the interrupt. Here's an example of legacy interrupts in firecracker: https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/device_manager/legacy.rs#L84 (it's outdated right now since it was using the old traits).

I think it's fine to keep the example out of this PR to increase focus, but there should be a matching device somewhere to understand how these interfaces are supposed to be used. For the pin-based interrupts everything is reasonably simple, I would still like to understand how the MSI interfaces are meant to be used though.

src/interrupt/legacy.rs
```diff
@@ -0,0 +1,39 @@
...
+    /// INTD
+    IntD = 0x4,
+}
+
+/// Standard configuration for Legacy interrupts.
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+pub struct LegacyIrqConfig {
+    /// Input of the system interrupt controllers the device's interrupt pin is connected to.
+    /// Implemented by any device that makes use of an interrupt pin.
+    pub interrupt_line: Option<u32>,
```
blitz, line 30: Would this be used by PCI devices to set their interrupt line field or are there other uses? Typically, the interrupt line field in the config space is just a normal R/W field that is written during boot by firmware. So device emulation should not be concerned with this value.

src/interrupt/mod.rs
alsrdn: Some controllers may have different registers to distinguish masked vs enabled interrupts. Take a look here: https://docs.microsoft.com/en-us/windows-hardware/drivers/gpio/gpio-interrupt-masks

There may also be functional differences in what happens when a masked interrupt is asserted. In some cases interrupt is not delivered when asserted until it is unmasked, whereas asserting a disabled interrupt should not keep the pending state.

src/interrupt/mod.rs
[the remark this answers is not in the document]
alsrdn: Take a look here: https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/device_manager/legacy.rs#L84

Also here: https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/builder.rs#L883

src/interrupt/legacy.rs
alsrdn: For a PCI device using an INTx interrupt interrupt_line may not be of significance for device emulation. But this value is of significance for setting up other components like the MPtable.
Also you would use `interrupt_line` to configure a specific line that you want the interrupt assigned to (see https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/device_manager/legacy.rs#L89).

alsrdn:
> > > Everything looks documented, but I still don't feel like I understand the usecases that drive the complexity.
> > > For devices that want to trigger MSIs, how does the interface look for them. Is there an example device somewhere?
> > 
> > 
> > I currently removed the example to keep this PR more focused following the discussion here [#71 (comment)](https://github.com/rust-vmm/vm-device/pull/71#discussion_r792849521). In order to trigger MSIs in a device emulation, configuration code should allocate, configure and enable the interrupts and pass an object with the EdgeInterrupt trait to the emulation code that will use `trigger()` to trigger the interrupt. Here's an example of legacy interrupts in firecracker: https://github.com/alsrdn/firecracker/blob/ab0c9642162f8c749c516d02b43b144a91baa33e/src/vmm/src/device_manager/legacy.rs#L84 (it's outdated right now since it was using the old traits).
> 
> I think it's fine to keep the example out of this PR to increase focus, but there should be a matching device somewhere to understand how these interfaces are supposed to be used. For the pin-based interrupts everything is reasonably simple, I would still like to understand how the MSI interfaces are meant to be used though.

For MSI it works the same way:

1. Create a MSI interrupt group in your VMM
2. Pass the group to the crate that is managing your device
3. Create interrupts within this group
4. Configure each interrupt
5. Enable the group if required
6. Pass each interrupt to the corresponding code emulating the queues
7. Trigger interrupts as needed

src/interrupt/msi.rs
blitz: The spec describes a 64-bit Message Address that is represented in MSI-X as hi/lo 32-bit parts. It's not really important, but it's weird to represent it here as 32-bit hi/lo parts as well, when you can just as well make it the 64-bit address that it is.

MSIs are normal data writes on the bus. So when they talk of "address" they literally mean an address where the device will write what's in the data field.

src/interrupt/mod.rs
blitz: This document describes behavior of a (GPIO) interrupt _controller_. So what you are saying is that the implementation of this trait needs to know what kind of interrupt controller it is connected to. This is fine, I'm just trying to wrap my head around things. :)

src/interrupt/mod.rs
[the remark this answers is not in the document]
blitz: This is from the virtio code you linked to:

```rust
    let mut interrupt_group = vmm.interrupt_manager.get_new_legacy_group().unwrap();
    interrupt_group.allocate_interrupts(1).unwrap();
    let irq = interrupt_group.get(0).unwrap();
    irq.update(&LegacyIrqConfig {
        interrupt_line: None,
        interrupt_pin: None,
    })
    .unwrap();
    let irq_config = irq.get_config().unwrap();
    interrupt_group.enable().unwrap();
```

This is frankly a terrible interface for legacy interrupt pins and should not end up in code that is intended for use in other VMMs. It has multi-step construction, it violates the "simple things should be simple" goal and the code is completely non-obvious:

- Why do you need to "allocate" after getting the "group" (for one interrupt?) from the manager?
- Why do need to "update" the group with basically no information right after creating it?
- Why do you need to `get_config` the config right after sticking the empty config into it?
- Why do you need to "enable" the interrupt?
- Why is it ok to unwrap all the results?

My advice here is to check out the interrupt setup in existing emulators/VMMs, for example [qemu](https://github.com/qemu/qemu/blob/master/hw/i386/pc_q35.c#L256). It is very readable  and this should especially be the goal for re-usable code.

Maybe the way forward is to merge this and iterate on it, but from my experience interfaces rarely become cleaner over time. I don't think I can contribute more here, unfortunately.

andreeaflorescu (approved):
At this point it's better to merge the PR, and iterate on the interfaces/responsibilities in a future release to unblock the future work.

